PR #12, opened 2018-05-11
Keep the translation after changing original string

adrianosferreira:
Different than single modules like Text in modules that contains many items, like Tabs, we were including a hash of the string content as string name, it was causing the translation to be lost whenever you change the original string.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5335

### DIFF
--- a/src/modules/class-wpml-elementor-module-with-items.php
+++ b/src/modules/class-wpml-elementor-module-with-items.php
@@ -38,7 +38,7 @@ abstract class WPML_Elementor_Module_With_Items implements IWPML_Page_Builders_M
 				if ( ! is_array( $field ) ) {
 					$strings[] = new WPML_PB_String(
 						$item[ $field ],
-						$this->get_string_name( $node_id, $item[ $field ], $field ),
+						$this->get_string_name( $node_id, $item[ $field ], $field, $element['widgetType'], $item['_id'] ),
 						$this->get_title( $field ),
 						$this->get_editor_type( $field )
 					);
@@ -46,7 +46,7 @@ abstract class WPML_Elementor_Module_With_Items implements IWPML_Page_Builders_M
 					foreach ( $field as $inner_field ) {
 						$strings[] = new WPML_PB_String(
 							$item[ $key ][ $inner_field ],
-							$this->get_string_name( $node_id, $item[ $key ][ $inner_field ], $inner_field ),
+							$this->get_string_name( $node_id, $item[ $key ][ $inner_field ], $inner_field, $item['_id'] ),
 							$this->get_title( $inner_field ),
 							$this->get_editor_type( $inner_field )
 						);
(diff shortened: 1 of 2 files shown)


src/modules/class-wpml-elementor-module-with-items.php
```diff
@@ -95,7 +95,7 @@ public function update( $node_id, $element, WPML_PB_String $string ) {
 	 * @return string
 	 */
 	private function get_string_name( $node_id, $value, $type, $key = '' ) {
-		return md5( $value ) . '-' . $type . $key . '-' . $node_id;
+		return $key . '-' . $type . '-' . $node_id;
```
strategio: @adrianosferreira, if you change the string name, doesn't this affect existing sites losing their translations?

brucepearson: @adrianosferreira Wont this cause backwards compatibility problems if you change the string name and old translations wont be found?

adrianosferreira:
@brucepearson @BruceOnTheGo please take a look.

src/modules/class-wpml-elementor-module-with-items.php
adrianosferreira: @BruceOnTheGo @brucepearson I don't think it can be a problem, because this method is only used in `\WPML_Elementor_Module_With_Items::update`. It will simply generate a new hash for the string name and put the translation on it.

brucepearson: @adrianosferreira It's also used in the get function - \WPML_Elementor_Module_With_Items::get.

Please run some tests to make sure that this is going to work properly with text that was translated before this change.

andreasneuber: Hi, did some testing with "old" elements and adjusted elements..
**Steps**
1. Create new test page
2. Click on blue Edit with Elementor button
3. Add some headings and text elements with "Old sth" (just to keep them apart better)
4. Add some elements with "Soon to be adjusted.." text
5. Save
6. Add translations via WPML translation editor, "Save & Close"
7. Then checkout WPML Core branch `wpmlcore-5335` and `composer update`
8. On test page, adjust the elements with "Soon to be adjusted.." to something else, leave all "Old ..." elements as they are
9. Go to Translation Editor again
10. Expected: All translations of "Old" elements stay as they are
11. Expected: Adjusted original text shows on left, adjusted translations show without a 2nd element being added

**Results**
- Though I also some had concerns as Bruce, with above steps "old" elements were not affected. They stayed intact.

- There is a small glitch with tab elements though:
-- When Tab element has only 1 tab => all good
-- When another tab is added, then in the TEditor the original text for tab 1 is replaced by text of new tab 2 and on right next to it we have translated text of tab 1 (I know, confusing :-)

---
1_old-elements-with-soon-to-adjusted-elements
2_new-tab-takes-over-first-tab

adrianosferreira: @andreasneuber hmm you found a bug. Let me double check it.

adrianosferreira: @andreasneuber @BruceOnTheGo @brucepearson I slightly changed the way we build the name of the string for modules which contains Items. I included the Item ID in the end of the name:

```
return $key . '-' . $type . '-' . $node_id . '-' . $item_id;
```

@andreasneuber could you double check it?

adrianosferreira:
@andreasneuber could you re-test it, please?
@BruceOnTheGo @brucepearson could you take a look once more?

brucepearson (dismissed):

strategio (dismissed):
@adrianosferreira, I sent a first comment and forgot to explicitly start the review so it was showing as pending... I had the same concern as Bruce, but since you tested it without any issue, it looks good to me. +1

PR #17, opened 2018-07-10
Adjust 'edit with Elementor' links when languages in domains

brucepearson:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5392

Please review @andreasciamanna @adrianosferreira @strategio @gordonsh24 @grey-rsi

### DIFF
--- a/src/class-wpml-elementor-urls.php
+++ b/src/class-wpml-elementor-urls.php
@@ -1,0 +1,35 @@
+<?php
+
+class WPML_Elementor_URLs implements IWPML_Action {
+
+	/** @var SitePress  */
+	private $sitepress;
+
+	public function __construct(
+		WPML_Translation_Element_Factory $element_factory,
+		WPML_Language_Domains $domains = null
+	) {
+		$this->element_factory = $element_factory;
+		$this->domains = $domains;
+	}
+
+	public function add_hooks() {
+		if ( $this->domains ) {
+			add_filter( 'elementor/document/urls/edit ', array( $this, 'adjust_edit_with_elementor_url' ), 10, 2 );
+		}
+	}
+
+	public function adjust_edit_with_elementor_url( $url, $elementor_document ) {
+		$post = $elementor_document->get_main_post();
+
+		$post_element = $this->element_factory->create( $post->ID, WPML_Translation_Element_Factory::ELEMENT_TYPE_POST );
+		$post_language = $post_element->get_language_code();
+
+		if ( $post_language ) {
+			$url_parts = wpml_parse_url( $url );
+			$url       = $url_parts['scheme'] . '://' . $this->domains->get( $post_language ) . $url_parts['path'] . '?' . $url_parts['query'];
+		}
+
+		return $url;
+	}
+}
--- a/src/class-wpml-elementor-urls.php
+++ b/src/class-wpml-elementor-urls.php
@@ -2,15 +2,12 @@
 
 class WPML_Elementor_URLs implements IWPML_Action {
 
-	/** @var SitePress  */
-	private $sitepress;
-
 	public function __construct(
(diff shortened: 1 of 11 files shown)


src/class-wpml-elementor-urls.php
```diff
@@ -0,0 +1,35 @@
...
+
+	public function adjust_edit_with_elementor_url( $url, $elementor_document ) {
+		$post = $elementor_document->get_main_post();
+
+		$post_element = $this->element_factory->create( $post->ID, WPML_Translation_Element_Factory::ELEMENT_TYPE_POST );
+		$post_language = $post_element->get_language_code();
+
+		if ( $post_language ) {
+			$url_parts = wpml_parse_url( $url );
+			$url       = $url_parts['scheme'] . '://' . $this->domains->get( $post_language ) . $url_parts['path'] . '?' . $url_parts['query'];
```
strategio: @brucepearson, I think you should use `http_build_url` instead of string concatenation. We can use it safely because we have a polyfill in WPML Core (https://github.com/jakeasmith/http_build_url).

src/class-wpml-elementor-urls.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+class WPML_Elementor_URLs implements IWPML_Action {
+
+	/** @var SitePress  */
+	private $sitepress;
```
strategio: You need to declare `$element_factory` and `$domains` (`$sitepress` is not use AFAICS).

src/class-wpml-elementor-urls.php
brucepearson: Thanks @strategio I knew there had to be are reverse for `parse_url` but I couldn't find :)

src/class-wpml-elementor-urls.php
brucepearson: Removed.

src/class-wpml-elementor-urls.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+class WPML_Elementor_URLs implements IWPML_Action {
+
```
strategio, line 4: @brucepearson, I thought you would also add the private properties `$element_factory` and `$domains`?

strategio (approved):
@brucepearson, except the private properties which I think should be declared, it looks good +1

src/class-wpml-elementor-urls.php
brucepearson: @strategio Thanks, I'm half asleep today.

tests/phpunit/stubs/stub-wpml-translation-element-factory.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+class WPML_Translation_Element_Factory {
```
andreasciamanna: @BruceOnTheGo all these stubs cause inspection problems (like "multiple definitions of {class_name}").

In this case, I think you can give it a more specific name, to avoid this issue, especially considering that in the future we may need something similar for other element types. 

Something like `WPML_Stub_Post_Element_Factory` should work.

andreasciamanna (changes requested):

tests/phpunit/stubs/stub-wpml-translation-element-factory.php
brucepearson: @andreasciamanna I needed the stub only for the class constant. I've refactored the WPML_Translation_Element_Factory in core so this is no longer needed.

andreasciamanna (approved):

tests/phpunit/stubs/stub-wpml-translation-element-factory.php
andreasciamanna: Cool. Thanks!

PR #24, opened 2018-07-18
[wpmlcore-5618] Check if variable is set before creating a String

adrianosferreira:
This fix is basically for CC. When you add Price List in a Elementor page, by default it always add "link" index in the JSON, even when it is empty. However, in the CC test it was not adding such index and throwing these notices.

### DIFF
--- a/src/modules/class-wpml-elementor-module-with-items.php
+++ b/src/modules/class-wpml-elementor-module-with-items.php
@@ -36,6 +36,11 @@ abstract class WPML_Elementor_Module_With_Items implements IWPML_Page_Builders_M
 		foreach ( $this->get_items( $element ) as $item ) {
 			foreach( $this->get_fields() as $key => $field ) {
 				if ( ! is_array( $field ) ) {
+
+					if ( ! isset( $item[ $field ] ) ) {
+						continue;
+					}
+
 					$strings[] = new WPML_PB_String(
 						$item[ $field ],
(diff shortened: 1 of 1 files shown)


andreasciamanna (commented):
@adrianosferreira if the issue is in the CC tests, why we should fix it in our code?

adrianosferreira:
@andreasciamanna as it was just a single check I didn't see any problem. @dimitrismitsis what do you think?

dimitrismitsis:
I agree and the fact is that this is possible programmatically when using Elementor, the problem isn't exactly on CC but on the way the post_meta are saved from wp-cli, CC actually isn't involved in this issue.  

Also, the fact that those PHP Notices aren't thrown from Elementor as well but only from wpml-page-builders indicate in my opinion that this should be added.

adrianosferreira:
@dimitrismitsis I'm afraid the problem is in the CC indeed, look:
https://git.onthegosystems.com/wpml/codeception/blob/master/_support/Helper/Elementor_Layout_Generator.php#L62

You are not setting the empty array for `link` index.

dimitrismitsis:
As discussed on MM, this is thrown from the native WP's function `update_post_meta` and I think it wouldn't hurt to include it.

andreasciamanna (approved):

andreasciamanna:
Ok then

PR #27, opened 2018-08-02
Replace image translation in the Image module

adrianosferreira:
This replacement happens when you access WPML Media Translation screen and add a translated image.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlmedia-511

### DIFF
--- a/src/class-wpml-elementor-media-translation.php
+++ b/src/class-wpml-elementor-media-translation.php
@@ -1,0 +1,65 @@
+<?php
+
+class WPML_Elementor_Media_Translation {
+
+	private $data_settings;
+	private $image_translate;
+	private $translation_element_factory;
+
+	public function __construct(
+		IWPML_Page_Builders_Data_Settings $data_settings,
+		WPML_Media_Image_Translate $image_translate,
+		WPML_Translation_Element_Factory $translation_element_factory
+	) {
+		$this->data_settings               = $data_settings;
+		$this->image_translate             = $image_translate;
+		$this->translation_element_factory = $translation_element_factory;
+	}
+
+	public function add_hooks() {
+		add_action( 'wpml_media_after_translate_media_in_post_content', array( $this, 'translate_image' ), PHP_INT_MAX, 3 );
+	}
+
+	public function translate_image( $post_id, $attachment_id, $target_language ) {
+		$post        = $this->translation_element_factory->create( $post_id, 'post' );
+		$translation = $post->get_translation( $target_language );
+
+		$attachment             = $this->translation_element_factory->create( $attachment_id, 'post' );
+		$attachment_translation = $attachment->get_translation( $target_language );
+
+		if ( $translation ) {
+			$custom_field_data = get_post_meta(
+				$translation->get_element_id(),
+				$this->data_settings->get_meta_field()
+			);
+
+			$custom_field_data = $this->data_settings->convert_data_to_array( $custom_field_data );
+			$this->update_images_in_modules(
+				$custom_field_data,
+				wp_get_attachment_url( $attachment_id ),
+				$post->get_language_code(),
+				$target_language,
+				$attachment_translation->get_element_id()
+			);
+
+			update_post_meta(
+				$translation->get_element_id(),
+				$this->data_settings->get_meta_field(),
+				$this->data_settings->prepare_data_for_saving( $custom_field_data )
+			);
+		}
+	}
+
+	private function update_images_in_modules( &$custom_field_data, $attachment_url, $source_language, $target_language, $attachment_translation_id ) {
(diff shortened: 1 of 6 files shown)


adrianosferreira:
@BruceOnTheGo @OnTheGoSystems/wpml please review it.

src/class-wpml-elementor-media-translation.php
```diff
@@ -0,0 +1,67 @@
...
+	public function add_hooks() {
+		add_action( 'wpml_media_after_translate_media_in_post_content', array( $this, 'translate_image' ), PHP_INT_MAX, 3 );
+	}
+
+	public function translate_image( $post_id, $attachment_id, $target_language ) {
+		$post        = $this->translation_element_factory->create( $post_id, 'post' );
+		$translation = $post->get_translation( $target_language );
+
+		$attachment             = $this->translation_element_factory->create( $attachment_id, 'post' );
+		$attachment_translation = $attachment->get_translation( $target_language );
```
brucepearson, line 28: @adrianosferreira You can move these two lines into the if statement because they are only needed then.

brucepearson (approved):
@adrianosferreira This looks ok. +1

src/class-wpml-elementor-media-translation.php
```diff
@@ -0,0 +1,67 @@
...
+
+			update_post_meta(
+				$translation->get_element_id(),
+				$this->data_settings->get_meta_field(),
+				$this->data_settings->prepare_data_for_saving( $custom_field_data )
+			);
+		}
+	}
+
+	private function update_images_in_modules( &$custom_field_data, $attachment_url, $source_language, $target_language, $attachment_translation_id ) {
```
strategio, line 53: Why do we need to pass `$custom_field_data` as a reference @adrianosferreira?

src/class-wpml-elementor-media-translation.php
```diff
@@ -0,0 +1,67 @@
...
+		foreach ( $custom_field_data as $key => &$data ) {
+			if ( is_array( $data ) ) {
+				$this->update_images_in_modules( $data, $attachment_url, $source_language, $target_language, $attachment_translation_id );
```
strategio, line 56: I think I understand @adrianosferreira. Would this work if you assign the result of the recursive call to `$data`?

`$data = $this->update_images_in_modules( $data, $attachment_url, $source_language, $target_language, $attachment_translation_id );`

strategio: And then remove the reference to the method arguments.

tests/phpunit/tests/test-wpml-elementor-media-translation-factory.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * Class Test_WPML_Elementor_Media_Translation_Factory
+ *
+ * @runTestsInSeparateProcesses
```
strategio, line 6: This might not be needed @adrianosferreira since the classes we overload are from another project.

strategio (commented):
@adrianosferreira, it sounds good but I think you should try to remove the argument passed as reference. Could you give a try to my suggestion?

src/class-wpml-elementor-media-translation.php
adrianosferreira: @strategio I see, your way should work too. I prefer to stick to the reference as we use the same way for handling other things, such as modules translation.

tests/phpunit/tests/test-wpml-elementor-media-translation-factory.php
adrianosferreira: Trying now...

adrianosferreira: @strategio no, I still need to run it in separated process, because in the second those overloaded classes shouldn't exist, I have this in the real code `if ( class_exists( 'WPML_Media_Image_Translate' ) && class_exists( 'WPML_Media_Attachment_By_URL_Factory' ) ) {`.

Makes sense?

src/class-wpml-elementor-media-translation.php
strategio: @adrianosferreira, this is confusing IMHO. We should choose either to pass as a reference and not return anything, or return something without passing it as a reference. I don't see any point to mix it.

I personally prefer avoiding arguments as a reference, but if you want to keep on this way, you should write:

```
$this->update_images_in_modules(
				$custom_field_data,
				wp_get_attachment_url( $attachment_id ),
				$post->get_language_code(),
				$target_language,
				$attachment_translation->get_element_id()
			);
```

and remove the return statement of `update_images_in_modules`.

strategio (approved):

PR #28, opened 2018-08-08
Handling page without translatable elements

adrianosferreira:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-146

### DIFF
--- a/tests/phpunit/tests/test-wpml-elementor-integration-factory.php
+++ b/tests/phpunit/tests/test-wpml-elementor-integration-factory.php
@@ -21,9 +21,10 @@ class Test_WPML_Elementor_Integration_Factory extends OTGS_TestCase {
 		$action_filter_loader->shouldReceive( 'load' )
 		                     ->once()
 		                     ->with( array(
-		                     	'WPML_Elementor_Translate_IDs_Factory',
-		                     	'WPML_Elementor_URLs_Factory',
+			                     'WPML_Elementor_Translate_IDs_Factory',
+			                     'WPML_Elementor_URLs_Factory',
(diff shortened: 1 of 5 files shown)


src/class-wpml-elementor-is-page-builder-page.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+class WPML_Elementor_Is_Page_Builder_Page {
```
brucepearson: @adrianosferreira I suspect this code will be exactly the same for Beaver Builder but only with a different meta field. It might be better if this code is in the page-builder repository and have a way for Elementor and Beaver Builder to supply the meta field that needs to be checked and copied.

adrianosferreira:
@BruceOnTheGo please review this also. We will do the same for Beaver Builder.

src/class-wpml-elementor-is-page-builder-page.php
adrianosferreira: @bruce done.

tests/phpunit/tests/test-wpml-elementor-integration-factory.php
```diff
@@ -23,6 +23,7 @@ public function it_creates() {
 		                     ->with( array(
 		                     	'WPML_Elementor_Translate_IDs_Factory',
 		                     	'WPML_Elementor_URLs_Factory',
+			                     'WPML_PB_Elementor_Handle_Custom_Fields_Factory',
```
strategio: Indentation glitch

strategio (dismissed):

brucepearson (dismissed):
@adrianosferreira Looks good. +1

PR #32, opened 2018-08-14
compsupp-3210- Setting HTML widget to AREA instead of Visual

amElnagdy:
Coming from https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-3210


amitkvint (approved):
looks good.

PR #33, opened 2018-08-15
Elementor WP Text Widget module integration

adrianosferreira:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5701


src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -47,24 +47,36 @@ public function get( $node_id, $element ) {
 		foreach ( $this->nodes_to_translate as $node_type => $node_data ) {
 			if ( $this->conditions_ok( $node_data, $element ) ) {
 				foreach ( $node_data['fields'] as $key => $field ) {
-					$field_key = $field['field'];
+					$field_key = isset( $field['field'] ) ? $field['field'] : '';
```
strategio, line 50: I would use `null` instead of an empty string for `$field_key` fallback value, but it's a personal taste.

strategio (approved):
It looks good @adrianosferreira +1

brucepearson (approved):
@adrianosferreira This looks good. +1

PR #39, opened 2018-09-10
Translate images in non-translatable blocks for Elementor

strategio:
I followed the same strategy as for Beaver Builder (https://github.com/OnTheGoSystems/wpml-page-builders-beaver-builder/pull/10 and https://github.com/OnTheGoSystems/wpml-page-builders/pull/24).

There's a lot of code (including the tests), but it does not make sense to split the task and it will be easier to run manual tests at once. Here are the most important parts:
- `WPML_Elementor_Media_Node_Provider` will provide node translators based on the type of element. It's easily extendable.
- `WPML_Elementor_Media_Nodes_Iterator` will iterate over all content nodes and will try to translate media nodes.
- `WPML_Elementor_Media_Node` is the base class for all the node translators.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5794

### DIFF
--- a/src/media/class-wpml-elementor-media-nodes-iterator.php
+++ b/src/media/class-wpml-elementor-media-nodes-iterator.php
@@ -1,0 +1,66 @@
+<?php
+
+class WPML_Elementor_Media_Nodes_Iterator implements IWPML_PB_Media_Nodes_Iterator {
+
+	/** @var WPML_Elementor_Media_Node_Provider $node_provider */
+	private $node_provider;
+
+	public function __construct( WPML_Elementor_Media_Node_Provider $node_provider ) {
+		$this->node_provider = $node_provider;
+	}
+
+	/**
+	 * @param array $data_array
+	 * @param string $lang
+	 * @param string $source_lang
+	 *
+	 * @return array
+	 */
+	public function translate( $data_array, $lang, $source_lang ) {
+		foreach ( $data_array as &$node ) {
+			if ( $this->is_parent_node( $node ) ) {
+				$node['elements'] = $this->translate( $node['elements'], $lang, $source_lang );
+			} elseif ( $this->is_valid_media_node( $node ) ) {
+				$node = $this->translate_node( $node, $lang, $source_lang );
+			}
+		}
+
+		return $data_array;
+	}
+
+	/**
+	 * @param array $node
+	 *
+	 * @return bool
+	 */
+	private function is_parent_node( $node ) {
+		return isset( $node['elements'] ) && $node['elements'];
+	}
+
+	/**
+	 * @param array $node
+	 *
+	 * @return bool
+	 */
+	private function is_valid_media_node( $node ) {
+		return isset( $node['elType'], $node['widgetType'], $node['settings'] )
+		       && 'widget' === $node['elType'];
+	}
+
+	/**
+	 * @param stdClass $node_data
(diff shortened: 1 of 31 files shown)


src/media/class-wpml-elementor-media-nodes-iterator.php
```diff
@@ -0,0 +1,49 @@
...
+	 *
+	 * @return array
+	 */
+	public function translate( $data_array, $lang, $source_lang ) {
+		foreach ( $data_array as &$node ) {
+			if ( isset( $node['elements'] ) && $node['elements'] ) {
+				$node['elements'] = $this->translate( $node['elements'], $lang, $source_lang );
+			} elseif ( isset( $node['elType'], $node['widgetType'], $node['settings'] )
+			           && 'widget' === $node['elType']
+			) {
```
andreasciamanna: @strategio perhaps you can extract the two expressions into two private methods:
- `hasElements` (the "$node" argument will give the context)
- `haselementType`

The second one, in particular, is a bit too long. The first one could be extracted for consistency.

andreasciamanna (approved):

src/media/class-wpml-elementor-media-nodes-iterator.php
strategio: Makes sense @andreasciamanna, extracted!

src/media/modules/class-wpml-elementor-media-node-image-gallery.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+class WPML_Elementor_Media_Node_Image_Gallery extends WPML_Elementor_Media_Node {
```
brucepearson: @strategio This class and the previous three are exactly the same except for the property name. Can you remove the code duplication.

src/media/modules/class-wpml-elementor-media-node-image.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+class WPML_Elementor_Media_Node_Image extends WPML_Elementor_Media_Node {
+
+	private $image_properties = array(
```
brucepearson: @strategio I don't think this is used.

src/media/modules/class-wpml-elementor-media-node-slides.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+class WPML_Elementor_Media_Node_Slides extends WPML_Elementor_Media_Node {
```
brucepearson: @strategio Slides and Carousel are exactly the same except for the property name. Please remove the duplication.

brucepearson (changes requested):
@strategio I've suggested some changes to remove the code duplication.

src/media/modules/class-wpml-elementor-media-node-image-gallery.php
strategio: Done @brucepearson, I created some abstract classes to reuse for nodes with the same pattern.

strategio:
@brucepearson, I fixed the duplicated code in the nodes, could you review it again please?

brucepearson:
> @brucepearson, I fixed the duplicated code in the nodes, could you review it again please?

GitHub was playing up when I tried to review this morning.

brucepearson (approved):

PR #40, opened 2018-09-12
Extend repeater fields from single to many

adrianosferreira:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5803

### DIFF
--- a/tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
+++ b/tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
@@ -385,4 +385,804 @@ class Test_WPML_Elementor_Translatable_Nodes extends OTGS_TestCase {
 
 		$this->assertEquals( $translation, $element['settings']['editor'] );
 	}
+
+	/**
+	 * @test
+	 * @group wpmlcore-5803
+	 */
+	public function it_gets_single_repeater_field() {
+		$subject = new WPML_Elementor_Translatable_Nodes();
+		$field_id = 1;
+
+		$nodes = array(
+			'my-custom-module'      => array(
+				'conditions'        => array( 'widgetType' => 'my-custom-module' ),
+				'fields'            => array(),
+				'integration-class' => 'WPML_PB_My_Custom_Module_With_A_Single_Repeater_Field',
+			),
+		);
+
+		WP_Mock::onFilter( 'wpml_elementor_widgets_to_translate' )
+		       ->with( $this->get_translatable_nodes() )
+		       ->reply( $nodes );
+
+		$element_data = array(
+			'widgetType' => 'my-custom-module',
+			'settings' => array(
+				'items' => array(
+					array(
+						'text' => 'my_text',
+						'_id' => $field_id,
+					),
+				),
+			),
+		);
+
+		$node_id = 123;
+
+		$string = new WPML_PB_String( 'my_text', 'my-custom-module-text-123-1', 'title', 'LINE' );
+
+		$this->assertEquals( array( $string ), $subject->get( $node_id, $element_data ) );
+	}
+
+	/**
+	 * @test
+	 * @group wpmlcore-5803
+	 */
+	public function it_gets_multiple_repeater_field() {
+		$subject = new WPML_Elementor_Translatable_Nodes();
+		$field_id = 1;
+
+		$nodes = array(
+			'my-custom-module'      => array(
+				'conditions'        => array( 'widgetType' => 'my-custom-module' ),
+				'fields'            => array(),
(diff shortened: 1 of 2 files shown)


adrianosferreira:
@OnTheGoSystems/wpml @BruceOnTheGo @brucepearson please review.

strategio (changes requested):
It looks good @adrianosferreira, but is there a way to cover this change in tests?

adrianosferreira:
@strategio they it is wrote now I don't think so. If you look, this part of the code "integration-class" didn't have tests before. I guess a CC test would be good.

strategio (approved):
@adrianosferreira, the changes are fine to me but I think we should start covering this code with test (either unit or CC, but I think it's possible to have unit tests).

adrianosferreira:
@strategio as these integration classes are not injected it is really difficult to write a single test for it. I can't see a way now. If you want (have time) to propose something it would be very helpful.

strategio:
> @strategio as these integration classes are not injected it is really difficult to write a single test for it. I can't see a way now. If you want (have time) to propose something it would be very helpful.

@adrianosferreira, I think we can hook on the filter `wpml_elementor_widgets_to_translate` to add integration classes (which would be an array of mocked classes), which is what a third party plugin would be doing I guess.

adrianosferreira:
> > @strategio as these integration classes are not injected it is really difficult to write a single test for it. I can't see a way now. If you want (have time) to propose something it would be very helpful.
> 
> @adrianosferreira, I think we can hook on the filter `wpml_elementor_widgets_to_translate` to add integration classes (which would be an array of mocked classes), which is what a third party plugin would be doing I guess.

@strategio so you are talking about integration tests or unit tests?

strategio:
Unit tests @adrianosferreira.

adrianosferreira:
> Unit tests @adrianosferreira.

@strategio what does the filter `wpml_elementor_widgets_to_translate` has to do with `WPML_Elementor_Translatable_Nodes`? Don't forget that we are trying to write an unit test for `\WPML_Elementor_Translatable_Nodes::get` ONLY for this part:

```php
if ( isset( $node_data['integration-class'] ) ) {
					foreach ( $this->get_integration_classes( $node_data ) as $class ) {
						try {
							$node    = new $class();
							$strings = $node->get( $node_id, $element, $strings );
						} catch ( Exception $e ) {
						}
					}
				}
```

strategio:
@adrianosferreira, you know better this code than me :)

When you call `get` or `update` methods, you first call the private method `initialize_nodes_to_translate` which itself applies the filter `wpml_elementor_widgets_to_translate`. So by hooking on this filter, you can add a node which has several integration classes.

I let you decide... I think it's possible to cover this but maybe I am missing something.

adrianosferreira:
> @adrianosferreira, you know better this code than me :)
> 
> When you call `get` or `update` methods, you first call the private method `initialize_nodes_to_translate` which itself applies the filter `wpml_elementor_widgets_to_translate`. So by hooking on this filter, you can add a node which has several integration classes.
> 
> I let you decide... I think it's possible to cover this but maybe I am missing something.

But we don't need to mock this filter, because in the translatable nodes we don't pass an instance of anything, we simply pass the class name as a plain string.

Actually, I think you have a point. I could mock this filter and return only nodes with integration tests for this single unit test, then I overload the class with Mockery in the `$node    = new $class();` part. I will check it out later today.

adrianosferreira:
@strategio I've managed to write tests to this.

tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
```diff
@@ -385,4 +385,804 @@ public function it_updates() {
...
+		$nodes = array(
+			'my-custom-module'      => array(
+				'conditions'        => array( 'widgetType' => 'my-custom-module' ),
+				'fields'            => array(),
+				'integration-class' => 'WPML_PB_My_Custom_Module_With_A_Single_Repeater_Field',
+			),
+		);
+
+		WP_Mock::onFilter( 'wpml_elementor_widgets_to_translate' )
+		       ->with( $this->get_translatable_nodes() )
```
strategio, line 406: @adrianosferreira, I am not sure if we should really pass all the nodes again (it's probably already covered in your other tests). Maybe you can use the code below (not tested...) as a suggestion.

```
->with( \WP_Mock\Functions::type( 'array' ) )->reply( $nodes )
```

strategio (approved):
Looks good to me @adrianosferreira. At least we have a basic coverage and we make sure we don't have problems outside the `try/catch` block.

tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
adrianosferreira: @strategio I didn't want to have these nodes, but unfortunately, I didn't find any alternative. Your idea doesn't work, the filter doesn't reply with my nodes.

andreasciamanna (approved):

PR #41, opened 2018-09-17
Only save post body as plain text if page has elementor data

brucepearson:
Only save post body as plain text if page has elementor data - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5841


andreasciamanna (approved):

strategio (approved):
Looks good @brucepearson. Don't we also have a similar problem with Beaver Builder?

adrianosferreira:
@strategio no, only Elementor hooks into `wpml_page_builder_string_translated` in order to save the post body.

PR #42, opened 2018-09-21
Force to use display as translated snippet for global widget in preview

strategio:
The preview is rendered in the admin so the display as translated snippet is not added to the query by default.

In parallel, we also need to set the CPT `elementor_library` to be displayed as translated (will do this in the remote config files).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5793

### DIFF
--- a/src/class-wpml-elementor-adjust-global-widget-id.php
+++ b/src/class-wpml-elementor-adjust-global-widget-id.php
@@ -27,6 +27,15 @@ class WPML_Elementor_Adjust_Global_Widget_ID {
 	public function add_hooks() {
 		add_action( 'elementor/editor/before_enqueue_scripts', array( $this, 'adjust_ids' ) );
 		add_action( 'elementor/editor/after_enqueue_scripts', array( $this, 'restore_current_language' ) );
+
+		if ( is_admin() ) {
+			add_filter(
+				'wpml_should_use_display_as_translated_snippet',
+				array( $this, 'should_use_display_as_translated_snippet' ),
+				PHP_INT_MAX,
+				2
+			);
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/class-wpml-elementor-adjust-global-widget-id.php
```diff
@@ -27,6 +27,15 @@ public function __construct(
 	public function add_hooks() {
 		add_action( 'elementor/editor/before_enqueue_scripts', array( $this, 'adjust_ids' ) );
 		add_action( 'elementor/editor/after_enqueue_scripts', array( $this, 'restore_current_language' ) );
+
+		if ( is_admin() ) {
+			add_filter(
+				'wpml_should_use_display_as_translated_snippet',
```
andreasciamanna, line 33: @strategio please use the same name as the function (+ prefix): `use_display_as_translated_snippet`.

However, what's "snippet"? The `use_display_as_translated_snippet` callback must contain this information in the PHPDoc. It must tell where this `$display_as_translated` is coming from. A developer should be able to track where the filter is called and I can't find this.

andreasciamanna (changes requested):

src/class-wpml-elementor-adjust-global-widget-id.php
strategio: @andreasciamanna, the filter `wpml_should_use_display_as_translated_snippet` already exists. So I renamed the callback to `should_use_display_as_translated_snippet`.

Also added a comment about the "snippet".

andreasciamanna (dismissed):

brucepearson (dismissed):

PR #45, opened 2018-10-15
Revert "Replace image translation in the Image module"

strategio:
This reverts commit 75e6ad9d839828a9ebf10d5b98fc051a2e5149aa (and goes with https://git.onthegosystems.com/wpml/wpml-media-translation/merge_requests/268).

This code was too specific and now we handle the whole media compatibility with `WPML_Elementor_Media_Hooks_Factory`.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5912


andreasciamanna (approved):
@strategio I'm not sure I understand this PR.

I see:

- you've added `WPML_Elementor_Media_Hooks_Factory` (which might be either a new class or a renamed `WPML_Elementor_Media_Translation_Factory`) to the list of dependencies in a test, but that's it
- You've removed four files

I would expect a more detailed explanation near the changed code of what I'm supposed to see here. 

Otherwise, I need to locally check out this branch + the one in WPML-Media, in order to (maybe) understand better.

Which is more or less what I did to get some idea and prevent wasting time, but in the future, please provide a bit more details: just looking at this PR and at the WPML-Media MR won't help.

PR #49, opened 2018-11-28
Replace the CSS class ID with the original for the global widget

strategio:
The CSS class `elementor-global-{translated_id}` will be replaced with `elementor-global-{original_id}`.

Note: Elementor is creating dynamic CSS files located in `wp-content/uploads/elementor/css`, but I didn't find an easy way to adjust the CSS class there.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6006

### DIFF
--- a/src/class-wpml-elementor-adjust-global-widget-id.php
+++ b/src/class-wpml-elementor-adjust-global-widget-id.php
@@ -27,6 +27,7 @@ class WPML_Elementor_Adjust_Global_Widget_ID {
 	public function add_hooks() {
 		add_action( 'elementor/editor/before_enqueue_scripts', array( $this, 'adjust_ids' ) );
 		add_action( 'elementor/editor/after_enqueue_scripts', array( $this, 'restore_current_language' ) );
+		add_action( 'elementor/frontend/the_content', array( $this, 'replace_css_class_id_with_original' ) );
 
 		if ( is_admin() ) {
 			add_filter(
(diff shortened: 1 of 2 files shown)


strategio:
@OnTheGoSystems/wpml review please!

kagg-design (dismissed):
+1 @strategio

src/class-wpml-elementor-adjust-global-widget-id.php
```diff
@@ -123,4 +124,20 @@ public function should_use_display_as_translated_snippet( $display_as_translated
...
+		$id      = (int) $matches[2];
+		$element = $this->translation_element_factory->create_post( $id );
+		$source  = $element->get_source_element();
+
+		if ( $source ) {
+			$before_id  = $matches[1];
+			$matches[0] = str_replace( $before_id . $id, $before_id . $source->get_id(), $matches[0] );
+		}
+
+		return $matches[0];
```
andreasciamanna, line 141: @strategio are we sure there are always all the 3 indexes here?

strategio: Yes @andreasciamanna, it's the `preg_replace` callback:
- `0` is for the whole regex pattern
- `1` is the first capturing group (before ID)
- `2` is the second capturing group (ID)

andreasciamanna: I know that, but what's in 1 and 2 when they don't capture anything? This what I don't know (thanks to my allergy to regular expressions :) )

strategio: If they don't capture the groups as defined in the regex, the callback should not be called.

andreasciamanna (dismissed):

dimitrismitsis:
@strategio tests are failing here

PR #52, opened 2018-11-29
Fix failing tests

strategio:
I the first commit I actually fixed the tests: Some changes were added in the node configuration and even though the
tests were changed, it has been lost (maybe a rebase).

See those commits:
- 890cc69310335ad5e2e678d4beef019ec14ca8b2
- 9792718623782d01b751146e3cda018b5c4d4f96
- 6cd73b728264c946d02f3b6babe1bae702e75b41

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6114

I added a second commit to move the node configuration to a public static function. To me, there's no point to test the configuration (that is, to copy the configuration into the tests).


kagg-design:
+1 @strategio

kagg-design (approved):
+1

strategio:
@OnTheGoSystems/wpml please check the second commit.

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -172,11 +173,10 @@ private function conditions_ok( $node_data, $element ) {
 		return $conditions_meet;
 	}
 
-	public function initialize_nodes_to_translate() {
-
-		$this->nodes_to_translate = array(
+	public static function get_nodes_to_translate() {
```
andreasciamanna, line 176: @strategio why changing it to static? We don't use static methods here

andreasciamanna (changes requested):

src/class-wpml-elementor-translatable-nodes.php
strategio: @andreasciamanna, the returned array does not depends on the state of the object. It's more a configuration array.

andreasciamanna: But all classes implementing `IWPML_Page_Builders_Translatable_Nodes` use non-satic method for `initialize_nodes_to_translate`.

The interface itself does not declare the method as static.

Declaring it static only in `class-wpml-elementor-translatable-nodes.php` not only makes the implementation inconsistent but violates the interface's contract.

strategio: @andreasciamanna, I didn't remove `initialize_nodes_to_translate`, it's still here https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/52/files#diff-1e5046e82faeba27c319819ef7a502caR673.

andreasciamanna: I've confused one method with the other. Sorry.

andreasciamanna (approved):

PR #54, opened 2018-12-10
Fix maintenance template query

adrianosferreira:
This fix is needed because WPML is calling `wp_reset_query` in `\SitePress::get_ls_languages` which happens after Elementor adjusted the query for displaying the coming soon template.

The `wp_reset_query` basically restores the global `wp_the_query`. So, we must adjust it before WPML calls the function.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5239

### DIFF
--- a/src/class-wpml-pb-fix-maintenance-query.php
+++ b/src/class-wpml-pb-fix-maintenance-query.php
@@ -1,0 +1,16 @@
+<?php
+
+class WPML_PB_Fix_Maintenance_Query {
+
+	const AFTER_TEMPLATE_APPLY = 12;
+
+	public function add_hooks() {
+		add_action( 'template_redirect', array( $this, 'fix_global_query' ), self::AFTER_TEMPLATE_APPLY );
+	}
+
+	public function fix_global_query() {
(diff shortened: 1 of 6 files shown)


adrianosferreira:
@OnTheGoSystems/wpml please review.

src/class-wpml-pb-fix-maintenance-query.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+class WPML_PB_Fix_Maintenance_Query {
+
+	public function add_hooks() {
+		add_action( 'template_redirect', array( $this, 'fix_global_query' ), 12 );
```
strategio: @adrianosferreira, could you put the `12` in a constant with a meaningful name?

strategio (changes requested):

adrianosferreira:
@strategio done.

strategio (dismissed):
+1 @adrianosferreira

brucepearson (dismissed):
@adrianosferreira +1

PR #55, opened 2018-12-27
Apply ELementor Maintenance fix when ST is not active

brucepearson:
The hook is activated by core so it works when ST is not active
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6196


ghost (approved):
@brucepearson looks good +1

PR #58, opened 2019-01-15
Elementor Price List description needs to be set to Area instead of Visual

Igor-WPML:
When you translate a price list widget by Elementor, an addition <p> tag will be added to the description due to the fact the description editor is set to Visual and it should be just Area.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6237


strategio (approved):
+1 @Igor-WPML

gordonsh24 (approved):

PR #62, opened 2019-02-04
Fixed PHP warning when dealing with _elementor_css field

adrianosferreira:
The problem happens because there is another indexed level inside of the field value (apparently changed recently on Elementor, not sure though).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6319

### DIFF
--- a/src/class-wpml-elementor-data-settings.php
+++ b/src/class-wpml-elementor-data-settings.php
@@ -35,7 +35,13 @@ class WPML_Elementor_Data_Settings implements IWPML_Page_Builders_Data_Settings 
 	 */
 	public function mark_css_field_as_empty( $value, $translated_post_id, $original_post_id, $meta_key ) {
 		if ( '_elementor_css' === $meta_key ) {
-			$value['status'] = '';
+			if ( ! isset( $value['status'] ) ) {
+				$value           = current( $value );
+				$value['status'] = '';
+				$value           = array( $value );
+			} else {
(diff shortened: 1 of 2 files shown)


adrianosferreira:
@OnTheGoSystems/wpml please review.

src/class-wpml-elementor-data-settings.php
```diff
@@ -35,7 +35,9 @@ public function add_hooks() {
 	 */
 	public function mark_css_field_as_empty( $value, $translated_post_id, $original_post_id, $meta_key ) {
 		if ( '_elementor_css' === $meta_key ) {
+			$value           = current( $value );
```
strategio: @adrianosferreira, if something changed inside Elementor, we will loose the backward compatibility. Can't you handle both cases?

Maybe:

```php
if ( ! isset( $value['status'] ) {
    $value           = current( $value );
    $value['status'] = '';
    $value           = array( $value );
} else {
    $value['status'] = ''
}
```

adrianosferreira: @strategio makes sense. I will change.

adrianosferreira: @strategio changed.

strategio (dismissed):
+1 @adrianosferreira

kagg-design (dismissed):
+1 @adrianosferreira

Issue #64, opened 2019-02-08
Multiple URL fields not working


synio-wesley:
Related to: https://github.com/pojome/elementor/issues/6786

We found out an issue where multiple URL fields cannot be translated.

Example code that is not working:

```
add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
    $widgets['example_name] = [
        'conditions' => ['widgetType' => 'example_name'],
        'fields' => [
            'url_field_1' => [
                'field' => 'url',
                'type' => __('Example URL Field 1', 'trailactive'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
            'url_field_2' => [
                'field' => 'url',
                'type' => __('Example URL Field 2', 'trailactive'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
        ],
    ];
    return $widgets;
});
```

I have found the issue to be inside of the `class-wpml-elementor-translatable-nodes.php`file.

To fix the issue, we can set `$field['field']` to something like `"{$key}_{$field_key}"` before calling the `get_string_name()` function in `get()` and `update()` (only when `!is_numeric($key)`). I have tested this change and it seems to work fine now!

**I can create a Pull Request if that would be helpful?**

adrianosferreira:
@synio-wesley yes, of course, we really appreciate that. Please create that and we will take a look.

strategio:
This has issue has been fixed by https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/67.

PR #65, opened 2019-02-08
Fix for issue #64 (Multiple URL fields not working)

synio-wesley:


adrianosferreira:
@synio-wesley thanks! Could you provide some guidelines to reproduce the problem?

synio-wesley:
@adrianosferreira I did my best with the unit test, but it seems like there's 1 error left that I don't understand... it seems to also be triggered without my changes though...?

```
There was 1 error:

1) Test_WPML_Elementor_Integration_Factory::it_creates
Mockery\Exception\NoMatchingExpectationException: No matching handler found for WPML_Action_Filter_Loader::load(array(0=>'WPML_Elementor_Translate_IDs_Factory',1=>'WPML_Elementor_URLs_Factory',2=>'WPML_Elementor_Adjust_Global_Widget_ID_Factory',3=>'WPML_PB_Elementor_Handle_Custom_Fields_Factory',4=>'WPML_Elementor_Media_Hooks_Factory',5=>'WPML_Elementor_WooCommerce_Hooks_Factory',)). Either the method was unexpected or its arguments matched no expected argument list for this method
```

Guidelines to reproduce the problem: Just try to translate a custom Elementor widget containing 2 URL fields like below. Without the fix, only one URL field can be translated because they will have the same name returned by the `get_string_name` function.

```
add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
    $widgets['example_name] = [
        'conditions' => ['widgetType' => 'example_name'],
        'fields' => [
            'url_field_1' => [
                'field' => 'url',
                'type' => __('Example URL Field 1', 'trailactive'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
            'url_field_2' => [
                'field' => 'url',
                'type' => __('Example URL Field 2', 'trailactive'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
        ],
    ];
    return $widgets;
});
```

Do you need a more complete example? Because I don't really have the time right now. Maybe I can do that in a few weeks..

synio-wesley:
@adrianosferreira I think the unit test was just not updated (yet) for one of the other changes that were introduced recently? I have modified test-wpml-elementor-integration-factory.php for that and now phpunit seems to give an OK.

Let me know if I need to change anything else before this can be merged.

synio-wesley:
Oh, it's not ready for production though. Because what would happen to the currently translated fields? You probably also need some sort of backwards compatibility..

Maybe we can instead simply add a custom key to override the default 'url' field name in the 'fields' configuration? So 'url' is still used internally, but the name of the field will be set to 'your-custom-field-name' when it is defined?

Not sure about the best approach here..

synio-wesley:
> Not sure about the best approach here..

Or perhaps it's possible to loop over all nodes on a next update of the plugin and then add the _$key_ prefix to the name for backwards compatibility where needed?.. (just thinking out loud)

strategio:
Hi @synio-wesley, sorry for our late reply.

I looked at the code, and I confirm that your change would cause a backward compatibility issue.

In order to have a better overview of the whole problem, could you share some guidelines on how to reproduce the problem? Are you using a custom widget? If yes, please share the code of this custom widget.

Thanks for you contribution :)

synio-wesley:
@strategio Thanks for the answer. I am using a custom widget. I'm registering the WPML support like this:

```
add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
    $widgets['example_name] = [
        'conditions' => ['widgetType' => 'example_name'],
        'fields' => [
            'url_field_1' => [
                'field' => 'url',
                'type' => __('Example URL Field 1', 'example-slug'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
            'url_field_2' => [
                'field' => 'url',
                'type' => __('Example URL Field 2', 'example-slug'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
        ],
    ];
    return $widgets;
});
```

I think the rest of the problem should be apparent. If you do need a more complete example, then I will need to create a new one, but I don't really have the time right now to be honest.

In any case, you can easily create any simple custom Elementor widget and then just use code similar to what I posted above (register 2 url fields) and you'll see that only one of them can be translated because they get assigned the same name internally.

strategio:
@synio-wesley, I understand the code passed to the filter `wpml_elementor_widgets_to_translate`, but could you share the code of your custom widget so we can use it as a reference in our tests?

synio-wesley:
@strategio Not right now, because I don't have it lying around. I created this example quickly when I was researching this problem for someone else. I'll make a note in my todo list to create a complete example for you somewhere after my vacation. Might take me a few weeks before I'll have some time for this though, because I have a huge backlog.

Or maybe @ryanlabelle can give you an example sooner. He is actually the one who first reported this issue. I will ask him.

strategio:
@synio-wesley, my question is why do you have the same `url` value for the `$field['field']`? Maybe I can understand better if I see the code of your custom widget.

What about the following:

```php
add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
    $widgets['example_name] = [
        'conditions' => ['widgetType' => 'example_name'],
        'fields' => [
            'url_field_1' => [
                'field' => 'url_field_1',
                'type' => __('Example URL Field 1', 'example-slug'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
            'url_field_2' => [
                'field' => 'url_field_2',
                'type' => __('Example URL Field 2', 'example-slug'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
        ],
    ];
    return $widgets;
});
```

Actually, if you look at the code base, you will see that we don't use keys for fields. We would even write this:

```php
add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
    $widgets['example_name] = [
        'conditions' => ['widgetType' => 'example_name'],
        'fields' => [
           [
                'field' => 'url_field_1',
                'type' => __('Example URL Field 1', 'example-slug'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
            [
                'field' => 'url_field_2',
                'type' => __('Example URL Field 2', 'example-slug'),
                'editor_type' => 'LINE', // or 'LINK' but relative links not possible that way
            ],
        ],
    ];
    return $widgets;
});
```

But of course, you will have to adjust your custom widget accordingly.

synio-wesley:
@strategio It would be cool if it would work like that, but it doesn't. And I actually had to read the source code to understand how to make it work (because I couldn't find it documented anywhere).

https://github.com/OnTheGoSystems/wpml-page-builders-elementor/blob/master/src/class-wpml-elementor-translatable-nodes.php line 53 checks whether key is numeric or not, and if not, it uses `$element[ $this->settings_field ][ $key ][ $field_key ]` instead of `$element[ $this->settings_field ][ $field_key ]` This is necessary, because `$element[ $this->settings_field ][ $field_key ]` would equal an array containing a field 'url' (among other fields).

This is why just using `'field' => 'url_field_1',` will not work here... It *does* work when you do it the way I do it, but then it only works for *just one* field, because the string name will be set to the same name for both fields (because `$key` is not being taken into account, only `$field_key`).

Perhaps one thing is missing from my example code there. This problem only happens if the field type in question is of type `\Elementor\Controls_Manager::URL`

If you still need a more complete example, then let me know and I'll cook something up after my holiday break.

synio-wesley:
More complete example code that should trigger the issue:

```
class Elementor_Example extends \Elementor\Widget_Base {
    public static $name = 'example';

    public static function prepare_wpml() {
        add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
            $widgets[self::$name] = [
                'conditions' => ['widgetType' => self::$name],
                'fields' => [
                    'example_url_1' => [
                        'field' => 'url',
                        'type' => __('Example URL 1', 'example'),
                        'editor_type' => 'LINE' // Or 'LINK' but then relative links won't work
                    ],
                    'example_url_2' => [
                        'field' => 'url',
                        'type' => __('Example URL 2', 'example'),
                        'editor_type' => 'LINE' // Or 'LINK' but then relative links won't work
                    ],
                ],
            ];
            return $widgets;
        });
    }

    public function get_name() {
        return self::$name;
    }

    public function get_title() {
        return __('Example', 'example');
    }

    public function get_icon() {
        return 'fa fa-heart';
    }

    public function get_categories() {
        return ['example'];
    }

    protected function _register_controls() {
        $this->start_controls_section('content_section', [
            'label' => __('Content', 'example'),
        ]);
        
        $this->add_control('example_url_1', [
            'label' => __('Example URL 1', 'example'),
            'type' => \Elementor\Controls_Manager::URL,
        ]);
        
        $this->add_control('example_url_2', [
            'label' => __('Example URL 2', 'example'),
            'type' => \Elementor\Controls_Manager::URL,
        ]);
    }

    protected function render() {
        $settings = $this->get_settings_for_display();
        echo $settings['example_url_1'] . ' / ' . $settings['example_url_2'];
    }

}

add_action('init', function() {
    Elementor_Example::prepare_wpml();
});
```

Note: just written this quickly as an example

One of the URL fields will be translatable now. You can also try to change the 'field' parameters to 'example_url_1' and 'example_url_2' instead of 'url' for both, but then none of both fields will be translatable.. which is even worse..

I'm bad at explaining, but I hope you understand what I'm trying to say?

If needed, I could do my best to write another PR taking backwards compatibility into account when I have some more time..

strategio:
Hi @synio-wesley, I suggest another way, simpler and backward compatible. Please look at https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/67.

Basically, you will have to pass an extra `field_id` for each field. So your map would look like this:

```php
public static function prepare_wpml() {
    add_filter('wpml_elementor_widgets_to_translate', function($widgets) {
        $widgets[self::$name] = [
            'conditions' => ['widgetType' => self::$name],
            'fields' => [
                'example_url_1' => [
                    'field'        => 'url',
                    'field_id'    => 'example_url_1', // New key
                    'type'        => __('Example URL 1', 'example'),
                    'editor_type' => 'LINE' // Or 'LINK' but then relative links won't work
                ],
                'example_url_2' => [
                    'field'        => 'url',
                    'field_id'    => 'example_url_2', // New key
                    'type'        => __('Example URL 2', 'example'),
                    'editor_type' => 'LINE' // Or 'LINK' but then relative links won't work
                ],
            ],
        ];
        return $widgets;
    });
}
```

Is that fine for you? Please feel free to test my branch https://github.com/OnTheGoSystems/wpml-page-builders-elementor/tree/pb-elementor-64 and let me know if it works as expected for you.

synio-wesley:
Hi @strategio, that sounds fine. I am not able to test it right now because I am on holiday now, but the code does look like it would fix the issue.

I will test out the branch somewhere next week after I am back. Thanks for looking into it.

Kind regards
Wesley Stessens

synio-wesley:
> Hi @strategio, that sounds fine. I am not able to test it right now because I am on holiday now, but the code does look like it would fix the issue.

Hi @strategio I'm not able to test this out soon, but thanks for the fix! Looking forward to being able to translate multiple URL fields in future projects!

strategio:
@synio-wesley, I am closing this PR because we handled it in https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/67. It will be introduced in the next WPML version (then we will update our [documentation](https://wpml.org/documentation/plugins-compatibility/elementor/how-to-add-wpml-support-to-custom-elementor-widgets/) accordingly)

PR #67, opened 2019-02-25
Introduce `field_id` to differenciate 2 occurrences of the same field in a node

strategio:
I started trying to dynamically detect multiple occurrences, and then automatically affect an extra suffix (except for the first one). But this was not working on the update part, because the strings will not necessarily be in the same order. Finally, I decided to introduce a new `field_wrapper` which will provide the actual name of the field to be used in the string.

This change will not affect existing strings and backward compatibility.

https://github.com/OnTheGoSystems/wpml-page-builders-elementor/issues/64
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6436

### DIFF
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -159,7 +159,8 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 	 * @return string
 	 */
 	public function get_string_name( $node_id, $field, $settings ) {
-		return $field['field'] . '-' . $settings[ self::TYPE ] . '-' . $node_id;
+		$field_id = isset( $field['field_id'] ) ? $field['field_id'] : $field['field'];
+		return $field_id . '-' . $settings[ self::TYPE ] . '-' . $node_id;
 	}
 
 	/**
--- a/tests/phpunit/tests/test-wpml-elementor-integration-factory.php
+++ b/tests/phpunit/tests/test-wpml-elementor-integration-factory.php
@@ -26,7 +26,7 @@ class Test_WPML_Elementor_Integration_Factory extends OTGS_TestCase {
 			                     'WPML_Elementor_Adjust_Global_Widget_ID_Factory',
 			                     'WPML_PB_Elementor_Handle_Custom_Fields_Factory',
 			                     'WPML_Elementor_Media_Hooks_Factory',
-								 'WPML_Elementor_WooCommerce_Hooks_Factory',
+			                     'WPML_Elementor_WooCommerce_Hooks_Factory',
 		                     ) );
 
 		$string_registration = \Mockery::mock( 'overload:WPML_PB_String_Registration' );
--- a/tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
+++ b/tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
@@ -575,6 +575,59 @@ class Test_WPML_Elementor_Translatable_Nodes extends OTGS_TestCase {
 
 	/**
 	 * @test
+	 * @group pb-elementor-64
+	 */
+	public function it_gets_multiple_occurrences_of_same_field_in_node() {
+		$subject = new WPML_Elementor_Translatable_Nodes();
+
+		$nodes = array(
+			'custom-module' => array(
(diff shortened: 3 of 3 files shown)


brucepearson (dismissed):
Don't forget to add a YT ticket for updating the documentation.
https://wpml.org/documentation/plugins-compatibility/elementor/how-to-add-wpml-support-to-custom-elementor-widgets/

strategio:
Right @brucepearson, this was in plans :) (note for myself https://onthegosystems.myjetbrains.com/youtrack/issue/wpmldoc-842).

kagg-design:
+1 @strategio

strategio:
I rebased the MR on https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/72.

strategio:
@BruceOnTheGo, I improved the code (after working on https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/67). I changed the naming and `field_id` became `field_wrapper` because it make more sense to me now.

Then I also used the `field_wrapper` key instead of the `$key` variable so we don't need to rely on this `$key` for the field definition.

The only field that is using the non-numerical key in our existing nodes is `link` (and some 3rd part code might use it too). This is still working, but we can't change it now without a migration. So I will not go further.

But all those changes will help to write a better documentation now.

Please review once again.

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -31,13 +31,14 @@ public function get( $node_id, $element ) {
 		foreach ( $this->nodes_to_translate as $node_type => $node_data ) {
 			if ( $this->conditions_ok( $node_data, $element ) ) {
 				foreach ( $node_data['fields'] as $key => $field ) {
-					$field_key    = $field['field'];
-					$string_value = null;
+					$field_key     = $field['field'];
+					$field_wrapper = isset( $field['field_wrapper'] ) ? $field['field_wrapper'] : $key;
```
brucepearson: @strategio What does `field_wrapper` really mean? Can you give an example? I suspect it's a bad name for it. I was also confused because Igor added `wrappers` for something else and I thought this was the same.

brucepearson (changes requested):

src/class-wpml-elementor-translatable-nodes.php
strategio: @BruceOnTheGo, basically we have 2 cases to handle:

1. String as a flat element:

```php
[
    'settings' => [
        'some_field' => 'This is a string to translate',
    ],
];
```

2. String "wrapped" inside an array:

```php
[
    'settings' => [
        'some_field' => [
            'something' => 'This is a string to translate',
        ],
    ],
];
```

As you can see in 2nd case, "some_field" could be considered as a wrapper of the actual field we want to translate "something".

Now, I understand your concern about the naming, this is something I missed... My inspiration is stuck, maybe we can call it `field_array`... Do you have some ideas?

brucepearson: @strategio How will the `$field` variable look for both these cases?

strategio: 1. Flat field

```php
[
    'field'       => 'some_field'
    'type'        => __( 'Some Field', 'sitepress' ),
    'editor_type' => 'LINE'
]
```

2. Array field

```php
[
    'field_wrapper' => 'some_field',
    'field'         => 'something',
    'type'          => __( 'Some Field', 'sitepress' ),
    'editor_type'   => 'LINE'
]
```

Note that in case 2, we will also keep the ability to use the array key as the "wrapper" for backward compatibility. We use it for "links" like this:

```php
'link' => [
	'field'       => 'url',
	'type'        => __( 'Image: Link URL', 'sitepress' ),
	'editor_type' => 'LINK'
]
```

brucepearson: So why don't we do this?
```php
'some_field' => [
    'field'         => 'something',
    'type'          => __( 'Some Field', 'sitepress' ),
    'editor_type'   => 'LINE'
]
```

strategio:
OK @brucepearson, back to the previous state (and rebased the branch), please vote again.

@OnTheGoSystems/wpml review please!

brucepearson (approved):

strategio:
@OnTheGoSystems/wpml, please give a second review here.

andreasciamanna (approved):

PR #69, opened 2019-03-18
Change Elementor Form required message id

amElnagdy:
It seems that Elementor changed the required message field id from `required_message  `to `required_field_message`

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6722

### DIFF
--- a/src/class-wpml-elementor-data-settings.php
+++ b/src/class-wpml-elementor-data-settings.php
@@ -34,7 +34,7 @@ class WPML_Elementor_Data_Settings implements IWPML_Page_Builders_Data_Settings 
 	 * @return mixed
 	 */
 	public function mark_css_field_as_empty( $value, $translated_post_id, $original_post_id, $meta_key ) {
-		if ( '_elementor_css' === $meta_key ) {
+		if ( '_elementor_css' === $meta_key && is_array( $value ) ) {
 			if ( ! isset( $value['status'] ) ) {
 				$value           = current( $value );
 				$value['status'] = '';
--- a/src/class-wpml-elementor-data-settings.php
+++ b/src/class-wpml-elementor-data-settings.php
@@ -35,7 +35,13 @@ class WPML_Elementor_Data_Settings implements IWPML_Page_Builders_Data_Settings 
 	 */
 	public function mark_css_field_as_empty( $value, $translated_post_id, $original_post_id, $meta_key ) {
 		if ( '_elementor_css' === $meta_key ) {
-			$value['status'] = '';
+			if ( ! isset( $value['status'] ) ) {
+				$value           = current( $value );
+				$value['status'] = '';
+				$value           = array( $value );
+			} else {
+				$value['status'] = '';
+			}
 		}
 
 		return $value;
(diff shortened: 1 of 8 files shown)


andreasciamanna (changes requested):
Won't this change break backwards compatibility?

I suggest to not replace the existing node, but rather add them as new nodes.

tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
```diff
@@ -339,7 +339,7 @@ public function node_data_provider() {
 					array( 'field' => 'email_from_name', 'type' => 'Form: Email from name', 'editor_type' => 'LINE' ),
 					array( 'field' => 'success_message', 'type' => 'Form: Success message', 'editor_type' => 'LINE' ),
 					array( 'field' => 'error_message', 'type' => 'Form: Error message', 'editor_type' => 'LINE' ),
-					array( 'field' => 'required_message', 'type' => 'Form: Required message', 'editor_type' => 'LINE' ),
```
andreasciamanna: Doesn't this also break backwards compatibility?

Shouldn't we have both `array( 'field' => 'required_message', 'type' => 'Form: Required message', 'editor_type' => 'LINE' ),` and `array( 'field' => 'required_field_message', 'type' => 'Form: Required message', 'editor_type' => 'LINE' ),`?

andreasciamanna (changes requested):

tests/phpunit/tests/test-wpml-elementor-translatable-nodes.php
amElnagdy: @andreasciamanna this is for tests and as far as I know, the tests are being run using the latest Elementor version which should not have the old required message field. Do I understand correctly?

andreasciamanna (approved):

andreasciamanna (changes requested):
Automated tests are failing

src/class-wpml-elementor-data-settings.php
```diff
@@ -34,8 +34,14 @@ public function add_hooks() {
...
-		if ( '_elementor_css' === $meta_key ) {
-			$value['status'] = '';
+		if ( '_elementor_css' === $meta_key && is_array( $value ) ) {
+			if ( ! isset( $value['status'] ) ) {
+				$value           = current( $value );
+				$value['status'] = '';
+				$value           = array( $value );
+			} else {
+				$value['status'] = '';
+			}
```
andreasciamanna: @amElnagdy how this part is related to the scope of the PR?

Also, this part is very confusing:

```php
$value           = current( $value );
$value['status'] = '';
$value           = array( $value );
```

Try to not reuse the same variable for the same purpose.
Use new variables instead, and give them a meaningful name, so the code will be understandable.

Something like that (but with better names, please):
```php
$current_value           = current( $value );
$current_value['status'] = '';
$value                   = array( $current_value );
```

I can't suggest any naming, because I really don't know why you are doing this, why you are setting the "status" key only to the "current" item of the `$value` array and then basically replace `$value` with a new array made of only one item.

andreasciamanna (changes requested):
In addition to the other comment, this PR requires a rebase.

src/class-wpml-elementor-data-settings.php
amElnagdy: @andreasciamanna Adriano authored this and I have no clue how this got into this PR.

andreasciamanna: @amElnagdy I think it might get solved if you rebase the branch.

PR #71, opened 2019-04-22
Improve code to not rely on numeric keys of the field definition 

strategio:
So far, we were assuming that numeric keys for the field definition were
matching with string value in the element field. And non-numeric keys
were matching with array fields. This was implicit but
not documented, and this is not correct IMO. We should rather check wether
the actual field is a string or an array instead.

The code is already covered with tests (so backward compatible) and I
added some more tests.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6442

I added a second commit (07cf0c5) to get rid of 2 private properties which are actually constants.


strategio:
@OnTheGoSystems/wpml, please check the fist commit independently (where we have the actual change).

strategio:
Closing in favor of https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/72 (with the correct branch name).

PR #37, opened 2018-08-28
Setting Flip Box Descriptions to AREA Type

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-3253

PR #72, opened 2019-04-22
Improve code to not rely on numeric keys of the field definition

strategio:
So far, we were assuming that numeric keys for the field definition were
matching with string value in the element field. And non-numeric keys
were matching with array fields. This was implicit but
not documented, and this is not correct IMO. We should rather check wether
the actual field is a string or an array instead.

The code is already covered with tests (so backward compatible) and I
added some more tests.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6442

I added a second commit (07cf0c5) to get rid of 2 private properties which are actually constants.

### DIFF
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -159,7 +159,8 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 	 * @return string
 	 */
 	public function get_string_name( $node_id, $field, $settings ) {
-		return $field['field'] . '-' . $settings[ self::TYPE ] . '-' . $node_id;
+		$field_id = isset( $field['field_id'] ) ? $field['field_id'] : $field['field'];
+		return $field_id . '-' . $settings[ self::TYPE ] . '-' . $node_id;
 	}
 
 	/**
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -10,27 +10,9 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 	const DEFAULT_HEADING_TAG = 'h2';
 
 	/**
-	 * @var string
-	 */
-	private $settings_field;
-
-	/**
-	 * @var string
-	 */
-	private $type;
-
-	/**
 	 * @var array
 	 */
 	private $nodes_to_translate;
-
-	/**
-	 * WPML_Elementor_Translatable_Nodes constructor.
-	 */
-	public function __construct() {
-		$this->settings_field = self::SETTINGS_FIELD;
-		$this->type           = 'widgetType';
-	}
 
 	/**
 	 * @param string|int $node_id Translatable node id.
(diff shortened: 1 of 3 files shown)


strategio:
@OnTheGoSystems/wpml, please check the fist commit independently (where we have the actual change).

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -49,28 +31,26 @@ public function get( $node_id, $element ) {
...
-							$this->get_string_name( $node_id, $field, $element ),
-							$field['type'],
-							$field['editor_type'],
-							$this->get_wrap_tag( $element )
-						);
-						$strings[] = $string;
-					} else if ( isset( $element[ $this->settings_field ][ $key ][ $field_key ] ) && trim( $element[ $this->settings_field ][ $key ][ $field_key ] ) ) {
-						$string    = new WPML_PB_String(
-							$element[ $this->settings_field ][ $key ][ $field_key ],
+					if ( isset( $element[ self::SETTINGS_FIELD ][ $field_key ] ) && is_string( $element[ self::SETTINGS_FIELD ][ $field_key ] ) && trim( $element[ self::SETTINGS_FIELD ][ $field_key ] ) ) {
```
andreasciamanna: @strategio I suggest you extract this into a private method which returns the result of `isset( $element[ self::SETTINGS_FIELD ][ $field_key ] ) && is_string( $element[ self::SETTINGS_FIELD ][ $field_key ] ) && trim( $element[ self::SETTINGS_FIELD ][ $field_key ] )`-

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -49,28 +31,26 @@ public function get( $node_id, $element ) {
...
+						$string_value = $element[ self::SETTINGS_FIELD ][ $field_key ];
+					} elseif ( isset( $element[ self::SETTINGS_FIELD ][ $key ][ $field_key ] ) && trim( $element[ self::SETTINGS_FIELD ][ $key ][ $field_key ] ) ) {
```
andreasciamanna: @strategio similar to the previous comment, you can extract this into a private method: it will help readability.

Also, consider what Bruce suggested some time ago (see https://github.com/tightenco/collect which allows you to use https://laravel.com/docs/5.8/collections).

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -106,10 +86,10 @@ public function update( $node_id, $element, WPML_PB_String $string ) {
 					$field_key = $field['field'];
 
 					if ( $this->get_string_name( $node_id, $field, $element ) === $string->get_name() ) {
-						if ( is_numeric( $key ) ) {
-							$element[ $this->settings_field ][ $field_key ] = $string->get_value();
+						if ( isset( $element[ self::SETTINGS_FIELD ][ $field_key ] ) && is_string( $element[ self::SETTINGS_FIELD ][ $field_key ] ) ) {
```
andreasciamanna: @strategio this looks almost the same as like #37, but is missing the `trim()`: is that in purpose?

If not, that would make my suggestion to extract the code even more important, because it could be reused.

Even if purposely omitted  `trim()`, you can still extract the code in two functions:
- One (let's call it "A") which returns the result of ` isset( $element[ self::SETTINGS_FIELD ][ $field_key ] ) && is_string( $element[ self::SETTINGS_FIELD ][ $field_key ] )`
- Another one (let's call it "B") which returns `A( $field_key ) && trim( $element[ self::SETTINGS_FIELD ][ $field_key ] )`

andreasciamanna (changes requested):

src/class-wpml-elementor-translatable-nodes.php
strategio: Right @andreasciamanna, this is extracted now.

src/class-wpml-elementor-translatable-nodes.php
strategio: @andreasciamanna, I think the `trim` check is not really needed when we update because we already had this checked when we created the string (if the string is empty, it won't be registered). However, I think it's better to have a symetric logic here to make it more readable and consistent. So I reused the same conditions (`is_flat_field` and `is_array_field`) and tests are still passing.

src/class-wpml-elementor-translatable-nodes.php
strategio: @andreasciamanna, I already looked at this library but I prefer to start using it on a new code.

andreasciamanna (dismissed):
As far as I can see, it looks ok now.

strategio:
@OnTheGoSystems/wpml one more review please :)

strategio:
@OnTheGoSystems/wpml, please give a second review here.

brucepearson (dismissed):
Nice work.

PR #73, opened 2019-04-24
make it possible to pass an instance of an integration class

lxbdr:
I created a reusable integration class for Elementor repeater fields where I need to add specifics by passing constructor params which is currently not possible. This addition makes it possible to pass an instance instead of a class string.

### DIFF
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -152,6 +152,13 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 	 * @return string
 	 */
 	public function get_string_name( $node_id, $field, $settings ) {
+		if ( 'heading' === $settings[ $this->type ] ) {
+			$header_size = isset( $settings[ $this->settings_field ]['header_size'] ) ? $settings[ $this->settings_field ]['header_size'] : '';
+			if ( $header_size ) {
+				return $field['field'] . '-' . $settings[ $this->type ] . '-' . $node_id . '-' . $header_size;
+			}
+		}
+
 		return $field['field'] . '-' . $settings[ $this->type ] . '-' . $node_id;
 	}
 
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -32,7 +32,7 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 	}
 
 	/**
-	 * @param string|int $node_id
+	 * @param string|int $node_id Translatable node id.
 	 * @param array $element
 	 *
 	 * @return WPML_PB_String[]
@@ -55,7 +55,8 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 							$element[ $this->settings_field ][ $field_key ],
 							$this->get_string_name( $node_id, $field, $element ),
 							$field['type'],
-							$field['editor_type']
+							$field['editor_type'],
+							$this->get_string_wrap( $element )
 						);
 						$strings[] = $string;
 					} else if ( isset( $element[ $this->settings_field ][ $key ][ $field_key ] ) && trim( $element[ $this->settings_field ][ $key ][ $field_key ] ) ) {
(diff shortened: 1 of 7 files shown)


src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -74,7 +74,11 @@ public function get( $node_id, $element ) {
 				if ( isset( $node_data['integration-class'] ) ) {
 					foreach ( $this->get_integration_classes( $node_data ) as $class ) {
 						try {
-							$node    = new $class();
+							if($class instanceof \WPML_Elementor_Module_With_Items){
```
strategio: @lxbdr, could you please add spaces around the parenthesis: `if ( $class instanceof \WPML_Elementor_Module_With_Items ) {`.

We are trying to follow the WP coding standards (https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style).

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -116,7 +120,11 @@ public function update( $node_id, $element, WPML_PB_String $string ) {
 				if ( isset( $node_data['integration-class'] ) ) {
 					foreach ( $this->get_integration_classes( $node_data ) as $class ) {
 						try {
-							$node = new $class();
+							if($class instanceof \WPML_Elementor_Module_With_Items){
```
strategio: Same here please @lxbdr.

strategio (changes requested):
Thanks for your contribution @lxbdr :)

I have a couple of formatting comments, but your change is legit in my opinion. Would you be able to cover this changes with a test? If not, no worries we will do it.

lxbdr:
@strategio thanks for the review. I added spaces to brackets. I tested the proposed changes on my setup and can confirm that it works.

Unfortunately I don't have any experience with unit tests so far but I looked at the existing code and tried to put something together. As I altered the method for getting and updating nodes I copied the tests for getting and updating single repeater fields so I think this should cover the changes - hope this is what you were looking for :)

strategio:
Thanks @lxbdr, and congratulations because your tests are passing :)

However, I see only partial formatting adjustments (and it's not "symetric"):
- `if( $class instanceof \WPML_Elementor_Module_With_Items ){`
- `if ( $class instanceof \WPML_Elementor_Module_With_Items ){`

Could you replace it by `if ( $class instanceof \WPML_Elementor_Module_With_Items ) {` ?

Meanwhile I will ask for another reviewer (@OnTheGoSystems/wpml).

strategio (approved):
It sounds good @lxbdr, thanks! I will ask for a second review and then I will ask you to squash your commits.

@OnTheGoSystems/wpml one more review please!

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -74,7 +74,11 @@ public function get( $node_id, $element ) {
 				if ( isset( $node_data['integration-class'] ) ) {
 					foreach ( $this->get_integration_classes( $node_data ) as $class ) {
 						try {
-							$node    = new $class();
+							if ( $class instanceof \WPML_Elementor_Module_With_Items ) {
+								$node = $class;
+							} else {
+								$node = new $class();
+							}
```
andreasciamanna, line 81: @strategio I'm not a huge fan of this sort of changes in existing API.

I do understand the reasons, but every time we do that, we make our API less deterministic.

Anyway, this is a month old PR: is it still valid?

lxbdr: Yes this PR is still valid.

I understand your argument and with checking whether it is an instance of the expected class it stays deterministic. I tried hard to find a different approach, with anonymous classes, using static properties, etc. but haven't found a way to make it work other than this...

strategio: Personnally, I think this solution is still valid. As an alternative, we could also allow to pass a factory, but it requires more changes I believe (and you won't be a huge fan of this either).

I'll leave you the final word.

andreasciamanna (approved):

andreasciamanna:
This PR still requires a refresh. Probably, just a rebase.

lxbdr:
@strategio just wanted to follow up on this one... Could you please do the 'squashing' or what is necessary so the PR can be merged?

strategio:
Hi @lxbdr! Thank you for reminding this PR, it was under the radar.

As I can't push anything on your repository, I created a squashed/rebased branch on ours, and opened a new PR (see https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/89).

dimitrismitsis:
This has been merged and it will be included in upcoming WPML bundle releases.

PR #75, opened 2019-06-03
Language Switcher Widget for Elementor

amElnagdy:
[I've been asked](https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6553) to create a language switcher widget for Elementor.
Thanks to @strategio's help, the widget is almost ready.

Note for me, more selectors still needed to cover various options in WPML->Languages -> custom language switchers

### DIFF
--- a/src/class-wpml-elementor-integration-factory.php
+++ b/src/class-wpml-elementor-integration-factory.php
@@ -20,7 +20,8 @@ class WPML_Elementor_Integration_Factory {
 				'WPML_Elementor_Adjust_Global_Widget_ID_Factory',
 				'WPML_PB_Elementor_Handle_Custom_Fields_Factory',
 				'WPML_Elementor_Media_Hooks_Factory',
-				'WPML_Elementor_WooCommerce_Hooks_Factory'
+				'WPML_Elementor_WooCommerce_Hooks_Factory',
+				\WPML\PB\Elementor\LanguageSwitcher\LanguageSwitcher::class,
 			)
 		);
 
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,20 @@
       "email": "hello@wpml.org"
     }
   ],
+  "repositories": [
+    {
+      "type":"package",
+      "package": {
+        "name": "elementor/elementor",
+        "version":"master",
+        "source": {
+          "url": "https://github.com/elementor/elementor.git",
+          "type": "git",
+          "reference":"master"
+        }
+      }
+    }
+  ],
   "autoload": {
     "classmap": [
       "src/"
@@ -27,6 +41,12 @@
   "require-dev": {
     "phpunit/phpunit": "~5.7",
     "otgs/unit-tests-framework": "~1.2.0",
-    "wpml/page-builders": "dev-master"
+    "wpml/page-builders": "dev-master",
+    "elementor/elementor": "dev-master"
+  },
+  "scripts": {
+    "post-package-install": [
+      "sed -i 's/final public function add_group_control/public function add_group_control/g' ./vendor/elementor/elementor/includes/base/controls-stack.php"
+    ]
   }
 }
--- a/src/LanguageSwitcher/LanguageSwitcher.php
+++ b/src/LanguageSwitcher/LanguageSwitcher.php
@@ -8,17 +8,10 @@ class LanguageSwitcher implements \IWPML_Backend_Action, \IWPML_Frontend_Action 
 
 	public function add_hooks() {
 		add_action( 'elementor/widgets/widgets_registered', [ $this, 'registerWidgets' ] );
-		add_filter( 'wpml_custom_language_switcher_is_enabled', [ $this, 'enableCustomLanguageSwitcher' ] );
+		add_filter( 'wpml_custom_language_switcher_is_enabled', '__return_true' );
 	}
 
 	public function registerWidgets() {
 		Plugin::instance()->widgets_manager->register_widget_type( new Widget() );
-	}
-
-	/**
-	 * @return bool
-	 */
-	public function enableCustomLanguageSwitcher() {
-		return true;
 	}
 }
--- a/src/LanguageSwitcher/LanguageSwitcher.php
+++ b/src/LanguageSwitcher/LanguageSwitcher.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace WPML\PB\Elementor\LanguageSwitcher;
+
+use Elementor\Plugin;
+
+class LanguageSwitcher implements \IWPML_Backend_Action, \IWPML_Frontend_Action {
+
+	public function add_hooks() {
+		add_action( 'elementor/widgets/widgets_registered', [ $this, 'registerWidgets' ] );
+	}
+
+	public function registerWidgets() {
+		Plugin::instance()->widgets_manager->register_widget_type( new Widget() );
+	}
+}
--- a/src/LanguageSwitcher/LanguageSwitcher.php
+++ b/src/LanguageSwitcher/LanguageSwitcher.php
@@ -8,9 +8,17 @@ class LanguageSwitcher implements \IWPML_Backend_Action, \IWPML_Frontend_Action 
 
 	public function add_hooks() {
 		add_action( 'elementor/widgets/widgets_registered', [ $this, 'registerWidgets' ] );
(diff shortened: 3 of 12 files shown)


src/language-switcher/class-wpml-elementor-language-switcher-factory.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
strategio: @amElnagdy, now that we use namespaces, you can use smaller file names => `factory.php` is enough :)

strategio (commented):
@amElnagdy, it sounds good. I have no very deep knowledge of the Elementor API to build custom widgets but what I see makes sense to me. It would be good if we could have some unit tests. Is this something you are planning to do?

src/class-wpml-elementor-integration-factory.php
```diff
@@ -20,7 +20,8 @@ public function create() {
 				'WPML_Elementor_Adjust_Global_Widget_ID_Factory',
 				'WPML_PB_Elementor_Handle_Custom_Fields_Factory',
 				'WPML_Elementor_Media_Hooks_Factory',
-				'WPML_Elementor_WooCommerce_Hooks_Factory'
+				'WPML_Elementor_WooCommerce_Hooks_Factory',
+				'\WPML\PB\Elementor\LanguageSwitcher\Factory'
```
brucepearson: @amElnagdy We no longer need factory classes for simple hooks now. You can change this to `'\WPML\PB\Elementor\LanguageSwitcher`. I'll add comments in other places about what else to change.

src/language-switcher/class-wpml-elementor-language-switcher-factory.php
```diff
@@ -0,0 +1,9 @@
...
+
+namespace WPML\PB\Elementor\LanguageSwitcher;
+
+class Factory implements \IWPML_Backend_Action_Loader, \IWPML_Frontend_Action_Loader {
```
brucepearson: @amElnagdy This factory wont be needed

src/language-switcher/class-wpml-elementor-language-switcher.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace WPML\PB\Elementor\LanguageSwitcher;
+
+class LanguageSwitcher implements \IWPML_Action {
```
brucepearson: @amElnagdy You can change this to implement `\IWPML_Backend_Action` and `\IWPML_Frontend_Action` and you wont need the factory.

brucepearson (changes requested):
@amElnagdy I've made some suggested changes. Can we have unit tests for all this?

src/language-switcher/class-wpml-elementor-language-switcher-factory.php
amElnagdy: Perfect, thanks @strategio!

As I have you here, to make this whole thing works properly the custom language switcher option in WPML -> Languages has to be enabled and the Dropdown option is selected. 
(As we already using wpml language switcher action)
So, should we enable it programmatically (is it possible?) or ask the clients to do?
What do you think?

More details: I can handle the different options in the custom language switcher (dropdown, dropdown click, horizontal list, etc..) by adding more selectors to the Elementor API but the optimal thing would just have a dropdown option as other options are easily passed to the action.

strategio: @amElnagdy, we should enable the custom language switcher programmatically. Please create a ticket in Core and assign it to me. I will provide some method to do this.

amElnagdy: Created - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6669

src/language-switcher/language-switcher.php
```diff
@@ -2,7 +2,7 @@
 
 namespace WPML\PB\Elementor\LanguageSwitcher;
 
-class LanguageSwitcher implements \IWPML_Action {
+class LanguageSwitcher implements \IWPML_Backend_Action, \IWPML_Frontend_Action {
```
amElnagdy: @brucepearson this is not loading the class, I'm getting a bunch of errors. Would you like to check? debug.log https://pastebin.com/myNeusth

brucepearson: @amElnagdy 
This is failing - class_implements(): Class \WPML\PB\Elementor\LanguageSwitcher\LanguageSwitcher
Did you run composer install to update autoloading?

amElnagdy: Not sure if I did. Currently, it's working fine. Thanks @brucepearson

andreasciamanna:
@amElnagdy can you please update the branch?

strategio:
I have rebased the branch and added some tests. Unfortunately, it's almost impossible to write tests because our widget is strongly coupled with Elementor's code (multiple inheritance) and Elementor's
code is strongly coupled (massive use of static methods and properties + singleton pattern).

I managed to cover the main `LanguageSwitcher` class by including the Elementor plugin from GitHub as a composer dependency. But I could not go further, covering the `Widget` class is impossible in pure unit, it would require integration tests.

@andreasciamanna, @brucepearson, @amElnagdy

andreasciamanna:
@amElnagdy or @strategio please add a reference to the ticket in the main commit message and in this PR.

Also please rearrange all the commits because at the moment it's not clear which one is a fixup and which one is a commit that must stay as-is.

Once this is done, make sure the ticket contains all the steps to test the widget, so we may add end-to-end tests.

andreasciamanna (dismissed):
It seems ok, but it's also a quite complex code. See my other comment in the PR.

strategio:
@amElnagdy, I'll let you squash/rebase the commits as you are the holder of this story.

src/language-switcher/widget/widget.php
```diff
@@ -0,0 +1,337 @@
+<?php
+namespace WPML\PB\Elementor\LanguageSwitcher;
+
+use Elementor\Controls_Manager;
+use Elementor\Group_Control_Typography;
+use Elementor\Scheme_Color;
+use Elementor\Widget_Base;
+
+class Widget extends Widget_Base {
```
brucepearson: @strategio If you really want to test the code properly then I think you need to remove all the business logic out of this class and use an adaptor pattern. The should make this `Widget` very simple so it doesn't have any logic that you need to test with unit tests. Then you can fully test the adaptor.

```php
__construct( Adaptor $adaptor ) {
   parent::__construct();
    $this->adaptor = $adaptor;
    $adaptor->target( $this );
}

public function get_name() {
    return $this->adaptor->getName();
}

...
...

protected function _register_controls() {
    $this->adaptor->registerControls();
}

```

Then in the Adaptor you call `$this->target->start_controls_section( .... )`, etc, which you can then test to confirm the target is called with the correct data.

strategio: Thanks for the suggestion @brucepearson. This is something I was thinking but I was unsure if all the methods used in `_register_controls` and `render` were public (I am talking of `get_settings_for_display`, `add_render_attribute`, `add_control`, ...). It looks like everything could be accessed by the adaptor. I will try to implement this!

strategio:
@brucepearson, I managed to implement the adaptor pattern but it was not so easy...

Some comments:
1. I decided to keep the elementor plugin as a dev package because we are
calling a lot of class constants which we would have to stub. I also
wanted to make sure that the methods we are using in the adaptor are
actually public (this is really not obvious).
2. One of the methods in the widget base class is
marked as "final" which does not allow mocking. I decided to remove this
when the package is installed in the dev dependencies (see script in composer.json).
3. Our `Widget` class can be instantiated by Elementor and so we can't
change the first 2 arguments in the constructor and we need to allow creating the
adaptor inside the constructor.

strategio:
The tests are failing in Scrutinizer because some Elementor classes could not be loaded. I can't find out the problem (did several blind tests), but things are working smoothly locally. I don't want to invest too much time on this since we are planning to move the repo to GitLab.

brucepearson (dismissed):
@strategio Nice work. +1

strategio:
I squashed all the commits because after the transposition into the "adatpor" pattern, it does not make sense do have different commits.

PR #83, opened 2019-07-08
Fix tests in PB Elementor

kagg-design:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6734


brucepearson (approved):
@kagg-design +1

kagg-design:
I think no second review is needed here.

PR #87, opened 2019-07-09
Change Form required message ID and adding missing fields

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6722


amElnagdy:
@OnTheGoSystems/wpml I would be grateful if you can spot where this test is failing.

brucepearson (dismissed):
@amElnagdy I fixed the test. It looks like the order of the fields is important. +1

andreasciamanna (dismissed):

PR #89, opened 2019-08-30
Make it possible to pass an instance of an integration class

strategio:
I created a reusable integration class for Elementor repeater fields
where I need to add specifics by passing constructor params which is
currently not possible. This addition makes it possible to pass an
instance instead of a class string.

Original PR:
https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/73

From: https://github.com/lxbdr


strategio:
@lxbdr FYI.

@dimitrismitsis, this PR is already reviewed in https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/73. Could you please handle it and merge it?

PR #92, opened 2019-10-01
Enable the custom language switcher option automatically

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6669


brucepearson (approved):

brucepearson (changes requested):
@strategio I think you should be using a filter rather than an action.

`
add_filter( 'wpml_use_custom_language_switcher', [ $this, 'someFIlter' ] );
`

strategio:
I will open another MR (the branch name is wrong).

PR #93, opened 2019-10-02
Enable the custom language switcher

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6669

### DIFF
--- a/src/LanguageSwitcher/LanguageSwitcher.php
+++ b/src/LanguageSwitcher/LanguageSwitcher.php
@@ -8,17 +8,10 @@ class LanguageSwitcher implements \IWPML_Backend_Action, \IWPML_Frontend_Action 
 
 	public function add_hooks() {
 		add_action( 'elementor/widgets/widgets_registered', [ $this, 'registerWidgets' ] );
-		add_filter( 'wpml_custom_language_switcher_is_enabled', [ $this, 'enableCustomLanguageSwitcher' ] );
+		add_filter( 'wpml_custom_language_switcher_is_enabled', '__return_true' );
 	}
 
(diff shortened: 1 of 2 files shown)


amElnagdy (approved):

src/LanguageSwitcher/LanguageSwitcher.php
```diff
@@ -8,9 +8,17 @@ class LanguageSwitcher implements \IWPML_Backend_Action, \IWPML_Frontend_Action
 
 	public function add_hooks() {
 		add_action( 'elementor/widgets/widgets_registered', [ $this, 'registerWidgets' ] );
+		add_filter( 'wpml_custom_language_switcher_is_enabled', [ $this, 'enableCustomLanguageSwitcher' ] );
```
dgwatkins: @strategio This is more a question than a review.

Is it ok to use the `__return_true` function here?

`add_filter( 'wpml_custom_language_switcher_is_enabled', '__return_true' );`

I usually do this and I am not sure if there is anything against it.

strategio: Yes @dgwatkins, this is totally valid. I forgot about this function.

But it also depends on the verbose degree of your code. Sometimes it can be useful to have an intermediary function to describe why we return `true`. But I don't think it applies here.

brucepearson: @dgwatkins @strategio The only problem with the `__return_true` approach is that it's not possible to remove with `remove_filter` but in this case I don't think it matters.

brucepearson (approved):

PR #94, opened 2019-10-03
Translate popups

strategio:
We will convert the popup ID on the fly as any other global element so
it does not depend on the order of the translations (the popup can be
translated after the page).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6542

### DIFF
--- a/src/Hooks/DynamicElements.php
+++ b/src/Hooks/DynamicElements.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+class DynamicElements implements \IWPML_Frontend_Action, \IWPML_DIC_Action {
+
+	public function add_hooks() {
+		add_filter( 'elementor/frontend/builder_content_data', [ $this, 'convert' ] );
+	}
+
+	/**
+	 * @param array $data
+	 *
+	 * @return array
+	 */
+	public function convert( array $data ) {
+		foreach ( $data as &$item ) {
+			if ( $this->isDynamicLink( $item ) ) {
+					$item['settings']['__dynamic__']['link'] = $this->convertPopUpId( $item['settings']['__dynamic__']['link'] );
+			}
+
+			$item['elements'] = $this->convert( $item['elements'] );
+		}
+
+		return $data;
+	}
+
+	/**
+	 * @param array $data
+	 *
+	 * @return bool
+	 */
+	private function isDynamicLink( array $data ) {
+		return isset( $data['elType'] )
+		       && 'widget' === $data['elType']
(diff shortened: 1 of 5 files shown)


strategio:
@OnTheGoSystems/wpml, please check this PR.

src/Hooks/DynamicElements.php
```diff
@@ -0,0 +1,74 @@
...
+		add_filter( 'elementor/frontend/builder_content_data', [ $this, 'convert' ] );
+	}
+
+	/**
+	 * @param array $dataArray
+	 *
+	 * @return array
+	 */
+	public function convert( array $dataArray ) {
+		foreach ( $dataArray as &$data ) {
```
andreasciamanna: @strategio we already know that the input is an array. Calling it `$dataArray` is quite redundant.

It seems like you called it like that because you were using `$data` to represent each item of the array.

If you change `$data` to `$item`, you can rename `$dataArray` to `$data`.

src/Hooks/DynamicElements.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+class DynamicElements implements \IWPML_Frontend_Action, \IWPML_DIC_Action {
```
brucepearson, line 5: @strategio I don't think you need `IWPML_DIC_Action` here because the class has no dependencies that I can see.

brucepearson (dismissed):
@strategio Looks good. +1

andreasciamanna:
I don't understand what I'm seeing, sorry :(

src/Hooks/DynamicElements.php
strategio: @andreasciamanna, this is the naming we use in other places of the code. But I agree that "data" and "item" is clearer. Changed!

src/Hooks/DynamicElements.php
strategio: @brucepearson, do you see a problem using the DIC instead of direct instantiation? Because I tend to always use the DIC even if the class does not have any dependency.

Now the class instantiation with the DIC is covered with an integration test and we don't need to care anymore about it if we decide to extend the class later (and introduce a dependency).

brucepearson: DIC will be a little slower but I don't mind either way. I just mentioned it just in case you added it by mistake or thought it was needed.

strategio:
I rebased the branch.

andreasciamanna (dismissed):

andreasciamanna:
@OnTheGoSystems/wpml please review

kagg-design (dismissed):
+1 @strategio

PR #99, opened 2019-11-18
Translate Divider widget text

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6950

### DIFF
--- a/src/modules/class-wpml-elementor-module-with-items.php
+++ b/src/modules/class-wpml-elementor-module-with-items.php
@@ -22,8 +22,10 @@ abstract class WPML_Elementor_Module_With_Items implements IWPML_Page_Builders_M
 	 */
 	abstract protected function get_editor_type( $field );
 
-	/** @return string */
-	abstract protected function get_items_field();
+	/**
+	 * @return string
+	 */
+	abstract public function get_items_field();
 
 	/**
 	 * @param string|int $node_id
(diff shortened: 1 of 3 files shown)


dgwatkins (dismissed):

strategio (dismissed):
+1 @amElnagdy

src/modules/class-wpml-elementor-module-with-items.php
```diff
@@ -22,8 +22,10 @@ abstract class WPML_Elementor_Module_With_Items implements IWPML_Page_Builders_M
 	 */
 	abstract protected function get_editor_type( $field );
 
-	/** @return string */
-	abstract protected function get_items_field();
+	/**
+	 * @return string
+	 */
+	abstract public function get_items_field();
```
andreasciamanna, line 28: @dgwatkins can you clarify this line? Where is this method implemented and used?

amElnagdy: @andreasciamanna it's used in src/class-wpml-elementor-translatable-nodes.php:110 
`$element[ self::SETTINGS_FIELD ][ $node->get_items_field() ][ $item['index'] ] = $item;`

strategio: @andreasciamanna, this abstract method was wrong since the beginning. All the classes implementing this interface are declaring a public method. I suggest @amElnagdy to fix it, but it has nothing to do with the purpose of this PR, so he added it as a separate commit.

andreasciamanna (dismissed):

PR #102, opened 2020-03-23
Support translation for dynamic content strings

strategio:
When a user selects a "dynamic source" for the content of an element, he
can also set 3 extra strings: "before", "after", "fallback".

These 3 strings will now be translatable.

Note: Dynamic content can be used for any kind of node (AFAICS) so it should apply to all nodes, even the ones that are not translatable.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6244

### DIFF
--- a/src/DynamicContent/Strings.php
+++ b/src/DynamicContent/Strings.php
@@ -1,0 +1,134 @@
+<?php
+
+namespace WPML\PB\Elementor\DynamicContent;
+
+use WPML\Collect\Support\Collection;
+use WPML_PB_String;
+
+class Strings {
+
+	const SETTINGS_REGEX        = '/settings="(.*?(?="]))/';
+	const NAME_PREFIX           = 'dynamic';
+	const DELIMITER             = '-';
+	const TRANSLATABLE_SETTINGS = [
+		'before',
+		'after',
+		'fallback',
+	];
+
+	/**
+	 * Remove the strings overwritten with dynamic content
+	 * and add the extra strings "before", "after" and "fallback".
+	 *
+	 * @param WPML_PB_String[] $strings
+	 * @param string            $nodeId
+	 * @param array             $element
+	 *
+	 * @return WPML_PB_String[]
+	 */
+	public static function filter( array $strings, $nodeId, array $element ) {
+
+		$dynamicFields = wpml_collect( isset( $element['settings']['__dynamic__'] ) ? $element['settings']['__dynamic__'] : [] );
+
+		$getStringKey = function ( WPML_PB_String $string ) {
+			return explode( '-', $string->get_name() )[0];
+		};
+
+		$updateFromDynamicFields = function( WPML_PB_String $string, $fieldKey ) use ( &$dynamicFields, $nodeId ) {
+			if ( $dynamicFields->has( $fieldKey ) ) {
+				return self::addBeforeAfterAndFallback( wpml_collect( [ $fieldKey => $dynamicFields->pull( $fieldKey ) ] ), $nodeId );
+			}
+
+			return $string;
+		};
+
+		return wpml_collect( $strings )
+			->keyBy( $getStringKey )
+			->map( $updateFromDynamicFields )
+			->merge( self::addBeforeAfterAndFallback( $dynamicFields, $nodeId ) )
+			->flatten()
+			->toArray();
+	}
+
+	private static function addBeforeAfterAndFallback( Collection $dynamicFields, $nodeId ) {
+		$dynamicFieldToSettingStrings = function( $dynamicTag, $tagKey ) use ( $nodeId ) {
+			preg_match( self::SETTINGS_REGEX, $dynamicTag, $matches );
+
+			$isTranslatableSetting = function( $value, $settingField ) {
+				return in_array( $settingField, self::TRANSLATABLE_SETTINGS );
+			};
+
+			$buildStringFromSetting = function( $value, $settingField ) use ( $nodeId, $tagKey ) {
+				return new WPML_PB_String(
+					$value,
+					self::getStringName( $nodeId, $tagKey, $settingField ),
+					sprintf( __( 'Dynamic content string: %s', 'sitepress' ), $tagKey ),
+					'LINE'
+				);
+			};
+
+			return wpml_collect( isset( $matches[1] ) ? self::decodeSettings( $matches[1] ) : [] )
+				->filter( $isTranslatableSetting )
+				->map( $buildStringFromSetting );
+		};
+		
+		return $dynamicFields->map( $dynamicFieldToSettingStrings );
+	}
+
+	/**
+	 * @param array          $element
+	 * @param WPML_PB_String $string
+	 *
+	 * @return array
+	 */
+	public static function updateNode( array $element, WPML_PB_String $string ) {
+		$stringNameParts = explode( self::DELIMITER, $string->get_name() );
+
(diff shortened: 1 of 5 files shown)


dgwatkins (commented):
@strategio I am paying attention to all functional PR but I don't think I am ready to review it yet.

strategio:
@dgwatkins, it's not fully functional programming yet, you can see some `if` blocks :)

Here I am targetting the `master` branch so I cannot use the new FP library, later we should be able to use only declarative code. But no worries, I am also far from this, I'll have to pratice it a lot...

@BruceOnTheGo, maybe you can give a review on this one.

@OnTheGoSystems/compatibility please review too.

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,142 @@
...
+	 * Remove the strings overwritten with dynamic content
+	 * and add the extra strings "before", "after" and "fallback".
+	 *
+	 * @param WPML_PB_String[] $strings
+	 * @param string            $nodeId
+	 * @param array             $element
+	 *
+	 * @return WPML_PB_String[]
+	 */
+	public static function filter( array $strings, $nodeId, array $element ) {
```
brucepearson, line 29: @strategio When you work with functional code you should try to get it to be more declarative. Basically this class filters the `$strings`. From the code comments:

> Remove the strings overwritten with dynamic content
	 * and add the extra strings "before", "after" and "fallback"

So I would expect the code to read something like:

```php
return wpml_collect( $strings )
   ->reject( $isOverwrittenWithDynamicContent )
   ->merge( $addBeforeAfterAndFallback( $dynamicFields ) )
   ->toArray();
```

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,142 @@
...
+		if ( ! isset( $element['settings']['__dynamic__'] ) ) {
+			return $strings;
+		}
+
+		$dynamicFields = wpml_collect( $element['settings']['__dynamic__'] );
```
brucepearson: @strategio I would combine these to one line:

```php
$dynamicFields = wpml_collect( isset( $element['settings']['__dynamic__'] ) ? $element['settings']['__dynamic__'] : [] );
```
I know it means a bit more processing rather than the quick exit of the function but I think the overall code is clearer and less cluttered.

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,142 @@
...
+
+		$isStaticStringOfDynamicField = function( WPML_PB_String $string ) use ( $dynamicFields ) {
```
brucepearson: @strategio Unless you have a real need to target master I would target develop and use the FP functions like curry and partial to make the code clearer.

```php
$isStaticStringOfDynamicField = curryN( 2, function( $dynamicFields, WPML_PB_String $string ) {
...
} );

...
->reject( $isStaticStringOfDynamicField( $dynamicFields )
....
```

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,142 @@
...
+						'LINE'
+					);
+				};
+
+				$strings = $strings->merge( wpml_collect( self::decodeSettings( $matches[1] ), true ) )
+					->filter( $isTranslatableSetting )
+					->map( $buildStringFromSetting );
+			}
+
+			return $strings;
```
brucepearson:
```php
return wpml_collect( isset( $matches[1] ) ? self::decodeSettings( $matches[1] ) : [] )
   ->filter( $isTranslatableSetting )
   ->map( $buildStringFromSetting );
```

brucepearson (commented):
@strategio I've made some suggestions. If you're code is fully covered with tests then you should be able to play with the code to get something that reads nice.

src/DynamicContent/Strings.php
strategio: That makes sense @brucepearson, I updated it in the way you suggested.

src/DynamicContent/Strings.php
strategio: @brucepearson, we decided to target `master` as much as we can because we have a lot of compatibility fixes to deliver. We might even request a compatibility specific update for Core (from `master`).

As soon as the FP library will be in `master`, we'll start using it.

src/DynamicContent/Strings.php
strategio: Done

strategio:
Thanks for your comments @brucepearson. The code is strongly covered with unit tests so I managed to apply most of your suggestions and make sure tests are still passing.

I just had to change the order of expected strings in a test. This is something that I missed but we should keep the the strings in the right order. A static string should be replaced with the 3 setting strings of the dynamic content. Do you have a recommended way to do this (if you already passed through this problem)?

brucepearson:
> I just had to change the order of expected strings in a test. This is something that I missed but we should keep the the strings in the right order. A static string should be replaced with the 3 setting strings of the dynamic content. Do you have a recommended way to do this (if you already passed through this problem)?

I'm not exactly sure what you need but maybe you need to create key for each existing string 'before', 'after', 'fallback' I assume. And then map to replace with new values and finally add any others that are missing.

strategio:
I improved the code (and tests) to make sure the dynamic content strings are located at the same position as the related static string (which is also removed). Then if we still have more dynamic content fields which don't originally have a static string, we'll had the remaining strings at the end of the collection.

@brucepearson please check again!

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @param string            $nodeId
+	 * @param array             $element
+	 *
+	 * @return WPML_PB_String[]
+	 */
+	public static function filter( array $strings, $nodeId, array $element ) {
+
+		$dynamicFields = wpml_collect( isset( $element['settings']['__dynamic__'] ) ? $element['settings']['__dynamic__'] : [] );
+
+		$setStringKey = function ( WPML_PB_String $string ) {
```
brucepearson: @strategio I think this should be called `$getStringKey`

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,136 @@
...
+			list( $fieldName ) = explode( '-', $string->get_name() );
```
brucepearson: @strategio I assume `list( $fieldName )` just gets the first element in the array. Why not:

```php
return explode( '-', $string->get_name() )[0];
```

src/DynamicContent/Strings.php
```diff
@@ -0,0 +1,136 @@
...
+
+			return $fieldName;
+		};
+
+		$replaceStaticStringWithSettingStrings = function( WPML_PB_String $string, $fieldKey ) use ( &$dynamicFields, $nodeId ) {
```
brucepearson: @strategio Maybe a shorter name - `$updateFromDynamicField`

brucepearson (dismissed):
@strategio Looks much nicer now. +1

strategio:
All done @brucepearson! I also added one commit to adjust missed tests for `WPML_Elementor_Translatable_Nodes::get`.

strategio:
BTW, I started running the unit test with Github Actions (https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/102/checks?check_run_id=533159403) but since yesterday it's failing to install `wpml/collect` (investigating) and we also have some tests that were already failing (will also investigate in https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7182).

PR #104, opened 2020-03-30
Fix failing tests in Elementor

strategio:
We had several problems and I kept every fix in a separate commit.

**FTR: We are loading Elementor plugin (free version) as a composer dependency because we use some constants from there.**

The commit that is affecting the code base is [Update deprecated classname `Scheme_Color` to `Core\Schemes\Color`](https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commit/3cd3250924197c390d3c004926b81144235e048b). The old class has been deprecated in Elementor `2.8.0` (released on 2019-12-09).

This failing test was caught because `ELEMENTOR_VERSION` constant is not defined in the tests and it triggered an error in `Elementor\Utils::handle_deprecation` (which is a good catch for us).

I tested `master` and this branch against Elementor `2.9.7` (current latest version) and I could not see any issue (the old class is an alias of the new one in Elementor's autoloader).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
     "wpml/collect": "dev-wpml-collect-rename"
   },
   "scripts": {
-    "post-install-cmd": [
+    "post-update-cmd": [
       "sed -i 's/final public function add_group_control/public function add_group_control/g' ./vendor/elementor/elementor/includes/base/controls-stack.php"
     ]
   }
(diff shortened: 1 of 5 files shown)


dgwatkins (approved):
:+1:

kagg-design (approved):
+1 @strategio

PR #105, opened 2020-04-04
Fix dynamic content strings when used for a module with items

strategio:
The current code was not supporting module with items (another data
shape) and it was completely removing the static strings from these.

To collect all the dynamic fields, I am using a collection of `Field`
objects which helps to match and replace a static string whether it's
for a simple module or module with items.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7194

### DIFF
--- a/src/DynamicContent/Strings.php
+++ b/src/DynamicContent/Strings.php
@@ -3,9 +3,15 @@
 namespace WPML\PB\Elementor\DynamicContent;
 
 use WPML\Collect\Support\Collection;
+use WPML_Elementor_Translatable_Nodes;
 use WPML_PB_String;
 
 class Strings {
+
+	const KEY_SETTINGS = WPML_Elementor_Translatable_Nodes::SETTINGS_FIELD;
+	const KEY_DYNAMIC  = '__dynamic__';
+	const KEY_NODE_ID  = 'id';
+	const KEY_ITEM_ID  = '_id';
 
 	const SETTINGS_REGEX        = '/settings="(.*?(?="]))/';
 	const NAME_PREFIX           = 'dynamic';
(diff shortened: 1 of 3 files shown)


src/DynamicContent/Strings.php
```diff
@@ -28,41 +34,112 @@ class Strings {
...
 	public static function filter( array $strings, $nodeId, array $element ) {
 
-		$dynamicFields = wpml_collect( isset( $element['settings']['__dynamic__'] ) ? $element['settings']['__dynamic__'] : [] );
+		$dynamicFields = self::getDynamicFields( $element );
 
-		$getStringKey = function ( WPML_PB_String $string ) {
-			return explode( '-', $string->get_name() )[0];
-		};
+		$updateFromDynamicFields = function( WPML_PB_String $string ) use ( &$dynamicFields, $nodeId ) {
+			$matchingField = $dynamicFields->first( function( Field $field, $key ) use ( $string ) {
```
strategio, line 40: Instead of looking at the string key (which was uncertain), we are now looking for a field object which is matching the static string. Then we replace the static string with the strings for the dynamic field.

src/DynamicContent/Strings.php
```diff
@@ -28,41 +34,112 @@ class Strings {
...
+				return isset( $item[ self::KEY_DYNAMIC ] );
+			} )
+			->map( function( array $item ) use ( $element ) {
+				return self::getFields(
+					$item[ self::KEY_DYNAMIC ],
+					$element[ self::KEY_NODE_ID ],
+					$item[ self::KEY_ITEM_ID ]
+				);
+			} )
+			->flatten();
```
brucepearson: @strategio In general I think it's easier to understand if the functions are extracted out:

```php
$isDynamic = function( array $item ) { return isset( $item[ self::KEY_DYNAMIC ] ); };
$getFields = function( array $item ) use ( $element ) {
				return self::getFields(
					$item[ self::KEY_DYNAMIC ],
					$element[ self::KEY_NODE_ID ],
					$item[ self::KEY_ITEM_ID ]
				);
			};

return wpml_collect( reset( $element[ self::KEY_SETTINGS ] ) )
			->filter( $isDynamic )
			->map( $getFields )
			->flatten();
```

I think it should be done here and elsewhere in this MR.

brucepearson (changes requested):

src/DynamicContent/Strings.php
strategio: Done @brucepearson! I also prefer to have named closures.

brucepearson (dismissed):
@strategio Looks good. +1

kagg-design (dismissed):
+1 @strategio

PR #107, opened 2020-04-15
Improve logic to replace CSS classes in templates and global elements

strategio:
I had to change the strategy and duplicate the CSS class with the
original object ID (instead of just replacing).

This is because Elementor is creating custom CSS stylesheets that are
targeting this original ID class name.

We will now also replace the class name "elementor-{translatedId}" to
"elementor-{translatedId} elementor-{originalId}".

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6630


dgwatkins (dismissed):
Looks good

dgwatkins:
@strategio Will you look into the "Inspection" check?

strategio:
> @strategio Will you look into the "Inspection" check?

Yes @dgwatkins, I'll do this when I have some time. The priority was to restore the pipeline for unit tests because it was broken and nobody was paying attention...

andreasciamanna (dismissed):

PR #108, opened 2020-05-14
Fix a fatal error when a module with items has strings instead of an array

strategio:
I don't know yet how to reproduce this, I believe it's because of a specific data shape. But I think removing the type hint is safe because this function will just check if it's an array with a specific element after.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7255

### DIFF
--- a/src/DynamicContent/Strings.php
+++ b/src/DynamicContent/Strings.php
@@ -79,7 +79,7 @@ class Strings {
 	 * @return Collection
 	 */
 	private static function getDynamicFieldsForModuleWithItems( array $element ) {
-		$isDynamic = function( array $item ) { return isset( $item[ self::KEY_DYNAMIC ] ); };
+		$isDynamic = function( $item ) { return isset( $item[ self::KEY_DYNAMIC ] ); };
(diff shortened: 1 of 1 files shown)


src/DynamicContent/Strings.php
```diff
@@ -79,7 +79,7 @@ private static function getDynamicFields( array $element ) {
 	 * @return Collection
 	 */
 	private static function getDynamicFieldsForModuleWithItems( array $element ) {
-		$isDynamic = function( array $item ) { return isset( $item[ self::KEY_DYNAMIC ] ); };
+		$isDynamic = function( $item ) { return isset( $item[ self::KEY_DYNAMIC ] ); };
```
andreasciamanna, line 82: @strategio I think a better approach would be to "reject", from the collection, the items which are not an array.

But this also works.

dgwatkins (approved):
Lets hope its enough :+1:

andreasciamanna (approved):
I don't want to hold the tagging because of my comment.

We can make it better later, as it won't change the behaviour of the fix.

PR #109, opened 2020-05-28
Support Form Steps in Elementor Pro 2.10.0

strategio:
I also added one commit to fix an missing array index. Not sure if it's an old or new problem. It's possible that the element keys are different from one version to another. We should fix it anyway...

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7275

### DIFF
--- a/src/modules/class-wpml-elementor-form.php
+++ b/src/modules/class-wpml-elementor-form.php
@@ -16,7 +16,17 @@ class WPML_Elementor_Form extends WPML_Elementor_Module_With_Items {
 	 * @return array
 	 */
 	public function get_fields() {
-		return array( 'field_label', 'placeholder', 'field_html', 'acceptance_text', 'field_options' );
+		return [
+			'field_label',
+			'placeholder',
+			'field_html',
+			'acceptance_text',
+			'field_options',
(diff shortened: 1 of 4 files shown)


dgwatkins (dismissed):
I'm pretty sure I reviewed this already 🤷

src/modules/class-wpml-elementor-form.php
```diff
@@ -16,7 +16,17 @@ public function get_items_field() {
...
+			'field_label',
+			'placeholder',
+			'field_html',
+			'acceptance_text',
+			'field_options',
+			'step_next_label',
+			'step_previous_label',
+			'previous_button',
+			'next_button',
+		];
```
andreasciamanna, line 29: @strategio when you mix refactoring with fixes, it's hard to tell one from the other.

andreasciamanna (dismissed):

src/modules/class-wpml-elementor-form.php
strategio: I understand @andreasciamanna. At least now it will be clearer the next time we alter the array elements.

PR #116, opened 2020-07-31
[wpmlcore-7279] Search widget fix with language as parameter

GeorgeBotsev:
[wpmlcore-7279] Add new class that is especially for Elementor search widget to handle different language search via hidden language parameter.
Add new class to be loaded on frontend actions

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7279

### DIFF
--- a/src/Hooks/Frontend.php
+++ b/src/Hooks/Frontend.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+class Frontend implements \IWPML_Frontend_Action {
+
+    public function add_hooks() {
+        add_action( 'elementor_pro/search_form/after_input', [ $this, 'addLanguageFormField' ] );
+    }
+
+    public function addLanguageFormField() {
+        do_action( 'wpml_add_language_form_field' );
+    }
+
+
+}
--- a/tests/phpunit/tests/Hooks/TestFrontend.php
+++ b/tests/phpunit/tests/Hooks/TestFrontend.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+/**
+ * @group hooks
+ * @group frontend
+ * @group wpmlcore-7279
+ */
+class TestFrontend extends \OTGS_TestCase {
+
+	/**
+	 * @test
+	 */
+	public function itShouldAddHooks() {
+		$subject = new Frontend();
+		\WP_Mock::expectActionAdded( 'elementor_pro/search_form/after_input', [ $subject, 'addLanguageFormField' ] );
+		$subject->add_hooks();
+	}
+
+	/**
+	 * @test
+	 */
+	public function itShouldAddLanguageFormField() {
+		$subject = new Frontend();
+		\WP_Mock::expectAction( 'wpml_add_language_form_field' );
+		$subject->addLanguageFormField();
+	}
+}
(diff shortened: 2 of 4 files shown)


src/Hooks/Frontend.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace WPML\PB\Elementor\Hooks;
+
+class Frontend implements \IWPML_Frontend_Action, \IWPML_DIC_Action {
+
+    public function add_hooks() {
+        add_action( 'elementor_pro/search_form/after_input', [ $this, 'elementor_add_language_form_field' ] );
+    }
+
+    public function elementor_add_language_form_field() {
```
strategio: @GeorgeBotsev, for new code (and when it's not forced by an interface), we want to use camelCase. Could you please rename if `addLanguageFormField`?

Note: We don't need to mention `elementor` in the method name, this is implicit.

tests/phpunit/tests/Hooks/TestFrontend.php
```diff
@@ -0,0 +1,20 @@
...
+
+	/**
+	 * @test
+	 */
+	public function TestAddHooks() {
+		$frontend = new Frontend();
+		\WP_Mock::expectActionAdded( 'elementor_pro/search_form/after_input', [ $frontend, 'elementor_add_language_form_field' ] );
+		$frontend->add_hooks();
+    }
+}
```
strategio, line 29: @GeorgeBotsev, if you want to cover the whole code (and we want this :), you also need a test on the method `addLanguageFormField`:

```php
/**
 * @test 
 */
public function itShouldAddLanguageFormField() {
     $subject = new Frontend();
    
    \WP_Mock::expectAction( 'wpml_add_language_form_field' );

    $subject->add_hooks();
}
```

tests/phpunit/tests/test-wpml-elementor-integration-factory.php
```diff
@@ -28,7 +28,8 @@ public function it_creates() {
 			                     'WPML_Elementor_Media_Hooks_Factory',
 			                     'WPML_Elementor_WooCommerce_Hooks_Factory',
 			                     \WPML\PB\Elementor\LanguageSwitcher\LanguageSwitcher::class,
-			                     \WPML\PB\Elementor\Hooks\DynamicElements::class,
+								 \WPML\PB\Elementor\Hooks\DynamicElements::class,
+								 \WPML\PB\Elementor\Hooks\Frontend::class,
```
strategio: There's something wrong with the indentation here @GeorgeBotsev.

strategio (changes requested):

tests/phpunit/tests/Hooks/TestFrontend.php
```diff
@@ -0,0 +1,29 @@
...
+		$subject->add_hooks();
+	}
+
+	/**
+	 * @test
+	 */
+	public function itShouldAddLanguageFormField() {
+		$subject = new Frontend();
+	   \WP_Mock::expectAction( 'wpml_add_language_form_field' );
+	   $subject->add_hooks();
```
strategio: @GeorgeBotsev, this is not correct (and tests are failing). You need to call the `addLanguageFormField` method:

```
$subject->addLanguageFormField();
```

Sorry, the example in my previous comment was wrong...

strategio: And please check the indentation.

strategio (dismissed):
+1 @GeorgeBotsev

src/Hooks/Frontend.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+class Frontend implements \IWPML_Frontend_Action, \IWPML_DIC_Action {
```
dgwatkins: @GeorgeBotsev I don't think we need `IWPML_DIC_Action` here because there is no constructor to inject dependencies to.

GeorgeBotsev: I was not sure about that.
I tested  now and it seems that you are right, so removing this now.

dgwatkins (dismissed):
Nice @GeorgeBotsev

strategio (dismissed):
Well done @GeorgeBotsev +1

dimitrismitsis:
@GeorgeBotsev This one could use some rebasing and conflict-resolving too!

PR #114, opened 2020-07-24
Add the ability to translate self-hosted video an external url for Elementor

GeorgeBotsev:
Adding the elements to our configuration, so that self-hosted video and external url video can be translated in the translation editor.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7154

PR #119, opened 2020-08-06
[wpmlcore-7439] Add the ability to translate reviews

GeorgeBotsev:
Add the ability to translate the reviews widget in Elementor

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7439

### DIFF
--- a/src/modules/Reviews.php
+++ b/src/modules/Reviews.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Reviews
+ */
+namespace WPML\PB\Elementor\Modules;
+
+class Reviews extends \WPML_Elementor_Module_With_Items {
+
+    protected function get_title( $field ) {
+        switch ( $field ) {
+            case 'content':
+                return esc_html__( 'Comment Contents', 'sitepress' );
+            case 'name':
+                return esc_html__( 'Commenter Name', 'sitepress' );
+            case 'title':
+                return esc_html__( 'Comment Title', 'sitepress' );
+            case 'image':
+                return esc_html__( 'Comment Image', 'sitepress' );
+            case 'url':
+                return esc_html__( 'Comment Link', 'sitepress' );
+            default:
+                return '';
+        }
+    }
+
+    public function get_fields() {
+        return [ 'content', 'name', 'title', 'link' => [ 'field' => 'url' ] ];
+    }
+
+    protected function get_editor_type( $field ) {
+        if ( 'content' === $field ) {
+            return 'LINE';
+        }
+        if ( 'name' === $field ) {
+            return 'LINE';
+        }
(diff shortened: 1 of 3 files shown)


src/modules/Reviews.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function get_fields()
+    {
+        return [ 'content', 'name', 'title', 'image' => [ 'url' ], 'link' => [ 'url' ] ];
+    }
+
+    protected function get_editor_type( $field )
+    {
+        if ( 'content' === $field ) {
+            return 'LINE';
```
amElnagdy, line 32: This should be AREA

src/modules/Reviews.php
```diff
@@ -0,0 +1,52 @@
...
+    protected function get_title( $field )
+    {
+        switch ( $field ) {
+            case 'content':
+                return esc_html__( 'Comment Contents', 'sitepress' );
+            case 'name':
+                return esc_html__( 'Commenter Name', 'sitepress' );
+            case 'title':
+                return esc_html__( 'Comment Title', 'sitepress' );
+            case 'image':
```
amElnagdy, line 17: Images are handled by Media Translation?

amElnagdy (changes requested):

src/modules/Reviews.php
GeorgeBotsev: Yes, they should be.
However, we already have the exception case here: https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/114 with self-hosted videos.
That is why I was trying to make it work in the same fashion.
Perhaps @strategio can give advice on what to do here.

src/modules/Reviews.php
GeorgeBotsev: Thats correct.
Thank you for pointing that out to me.

src/modules/Reviews.php
GeorgeBotsev: I looked further at this.
It is a different case than the self-hosted video.
In the self-hosted video you can specify URL, while here you can only use the media gallery to set the attachment.
I am going to remove the ability to manually specify the image path because of that.

amElnagdy (dismissed):
Looks good 👍

src/modules/Reviews.php
strategio: If we want to translate the image, we need to create a new media module in `src/media/modules` with a class extending `\WPML_Elementor_Media_Node`.

I am sharing for you to know, but I would not bother implement this here. The image should be a picture of the reviewer, so I don't see why it would be converted.

strategio (dismissed):
+1 @GeorgeBotsev

PR #120, opened 2020-08-07
Fixed replacing translated media in posts when using ATE

mrudzki91:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7166

wpmlcore-7166


strategio (dismissed):
+1 @mrudzki91 

Small note: I suggest to use the full url to the ticket in the commit message, that way we can click it :)

mrudzki91:
> +1 @mrudzki91
> 
> Small note: I suggest to use the full url to the ticket in the commit message, that way we can click it :)

Sorry for that - I was used to GitLab where ticket number is automatically linked to the URL. :)

brucepearson (dismissed):

PR #121, opened 2020-08-07
Prevent mix of string packages if a post was initially edited with Gutenberg and then with Elementor

strategio:
When Elementor is saving its data, we'll try to:
- detect some Gutenberg block meta data in all editor widgets and remove it
- if Elementor's data has changed, update the data
- remove the Gutenberg package

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7444

### DIFF
--- a/src/DataConvert.php
+++ b/src/DataConvert.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace WPML\PB\Elementor;
+
+class DataConvert {
+
+	/**
+	 * @param array $data
+	 *
+	 * @return string
+	 */
+	public static function serialize( array $data ) {
+		return wp_slash( wp_json_encode( $data ) );
+	}
+
+	/**
+	 * @param array|string $data
+	 *
+	 * @return array
+	 */
+	public static function unserialize( $data ) {
+		return json_decode( is_array( $data ) ? $data[0] : $data, true );
+	}
+}
--- a/src/Hooks/GutenbergCleanup.php
+++ b/src/Hooks/GutenbergCleanup.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+use WPML\FP\Fns;
+use WPML\FP\Logic;
+use WPML\FP\Maybe;
+use WPML\FP\Relation;
+use WPML\LIB\WP\Post;
+use WPML\PB\Elementor\DataConvert;
+use WPML_Elementor_Data_Settings;
+use function WPML\FP\curryN;
+use function WPML\FP\pipe;
+
+class GutenbergCleanup implements \IWPML_Backend_Action, \IWPML_Frontend_Action {
+
+	public function add_hooks() {
+		add_filter(
+			'update_post_metadata',
+			Fns::withoutRecursion( Fns::identity(), [ $this, 'removeGutenbergFootprint' ] ),
+			10,
+			4
+		);
+	}
+
+	/**
+	 * If we detect Gutenberg footprint in the Elementor data,
+	 * we'll remove it and delete the Gutenberg string package.
+	 *
+	 * @param null|bool $check
+	 * @param int       $postId
+	 * @param string    $metaKey
+	 * @param string    $metaValue
+	 *
+	 * @return mixed
+	 */
+	public function removeGutenbergFootprint( $check, $postId, $metaKey, $metaValue ) {
+		if (
+			WPML_Elementor_Data_Settings::META_KEY_DATA === $metaKey
+			&& WPML_Elementor_Data_Settings::is_edited_with_elementor( $postId )
+		) {
+			// $ifValueHasChanged :: string -> bool
+			$ifValueHasChanged = pipe( Relation::equals( $metaValue ), Logic::not() );
+
+			// $update :: int -> string -> bool
+			$update = curryN( 2, function( $postId, $meta ) {
+				Post::updateMeta( $postId, WPML_Elementor_Data_Settings::META_KEY_DATA, $meta );
+				self::deletePackage( self::getGutenbergPackage( $postId ) );
+				return true;
+			} );
+
+			return Maybe::of( $metaValue )
+				->map( [ DataConvert::class, 'unserialize' ] )
+				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
+				->map( [ DataConvert::class, 'serialize' ] )
+				->filter( $ifValueHasChanged )
+				->map( $update( $postId ) )
+				->getOrElse( $check );
+		}
+
+		return $check;
+	}
+
+	/**
+	 * @param array $data
+	 *
+	 * @return array
+	 */
+	public function removeBlockMetaInEditorWidget( array $data ) {
+		foreach ( $data as &$element ) {
+			if ( $element['elements'] ) {
+				$element['elements'] = $this->removeBlockMetaInEditorWidget( $element['elements'] );
+			} elseif ( 'widget' === $element['elType'] && isset( $element['settings']['editor'] ) ) {
+				$element['settings']['editor'] = preg_replace( '(<!--[^<]*-->)', '', $element['settings']['editor'] );
+			}
+		}
+
+		return $data;
+	}
+
+	/**
+	 * @param int $postId
+	 */
+	public static function getGutenbergPackage( $postId ) {
+		// $isGbPackage :: \WPML_Package -> bool
+		$isGbPackage = Relation::propEq( 'kind_slug', 'gutenberg' );
+
+		return wpml_collect( apply_filters( 'wpml_st_get_post_string_packages', [], $postId ) )
+			->filter( $isGbPackage )
+			->first();
+	}
+
+	/**
+	 * @param \WPML_Package|null $package
+	 */
+	public static function deletePackage( $package ) {
+		$package && do_action( 'wpml_delete_package', $package->name, $package->kind );
+	}
+}
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,10 @@
     "otgs/unit-tests-framework": "~1.2.0",
     "wpml/page-builders": "dev-master",
     "elementor/elementor": "dev-master",
-    "wpml/collect": "dev-wpml-collect-rename"
+    "wpml/collect": "dev-wpml-collect-rename",
+    "wpml/fp": "^0.1.1",
+    "wpml/wp": "^0.1.1",
+    "composer/composer": "^1.10"
   },
   "scripts": {
     "post-update-cmd": [
--- a/src/class-wpml-elementor-data-settings.php
+++ b/src/class-wpml-elementor-data-settings.php
@@ -1,6 +1,11 @@
 <?php
 
+use WPML\PB\Elementor\DataConvert;
+
 class WPML_Elementor_Data_Settings implements IWPML_Page_Builders_Data_Settings {
+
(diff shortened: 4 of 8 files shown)


strategio:
@OnTheGoSystems/compatibility, @OnTheGoSystems/wpml please check this.

src/Hooks/GutenbergCleanup.php
```diff
@@ -0,0 +1,108 @@
...
+use WPML\FP\Fns;
+use WPML\FP\Maybe;
+use WPML\FP\Obj;
+use WPML\FP\Relation;
+use WPML\PB\Elementor\DataConvert;
+use WPML_Elementor_Data_Settings;
+use function WPML\FP\partial;
+use function WPML\FP\pipe;
+
+class GutenbergCleanup implements \IWPML_Backend_Action, \IWPML_Frontend_Action, \IWPML_DIC_Action {
```
dgwatkins: @strategio why do we need `IWPML_DIC_Action` if there is no constructor?

src/Hooks/GutenbergCleanup.php
```diff
@@ -0,0 +1,108 @@
...
+
+	public function add_hooks() {
+		add_filter(
+			'update_post_metadata',
+			Fns::withoutRecursion( Fns::identity(), [ $this, 'removeGutenbergFootprint' ] ),
```
dgwatkins, line 20: @strategio can you explain `Fns::identity()` ?

I've always used `Fns::noop()` there

dgwatkins (commented):
@strategio just a couple of questions.

src/Hooks/GutenbergCleanup.php
strategio: @dgwatkins, if we are in a recursion `Fns::withoutRecursion` will call the first function, `Fns::identify` here, so it will return the original value passed to the filter.

At some point, it's like replacing our current filter with a filter that just returns the value it receives.

If I used `Fns::noop`, it would be a filter that always returns `null` (because does not return anything), regardless of the input value.

Did I missed something?

dgwatkins: Makes sense, thanks!

src/Hooks/GutenbergCleanup.php
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	 * @param int $postId
+	 */
+	public static function getGutenbergPackage( $postId ) {
+		// $isGbPackage :: \WPML_Package -> bool
+		$isGbPackage = pipe( Obj::prop( 'kind_slug' ), Relation::equals( 'gutenberg' ) );
+
+		return wpml_collect( apply_filters( 'wpml_st_get_post_string_packages', [], $postId ) )
+			->filter( $isGbPackage )
```
brucepearson, line 89: @strategio You can use Relation::propEq here.
```php
->filter( Relation::propEq( 'kind_slug', 'gutenberg' ) )
```

src/Hooks/GutenbergCleanup.php
strategio: Yes, this is definitely optional @dgwatkins, but it should not hurt :)

src/DataConvert.php
```diff
@@ -0,0 +1,29 @@
...
+	 *
+	 * @return array
+	 */
+	public static function unserialize( $data ) {
+		$convertedData = $data;
+		if ( is_array( $data ) ) {
+			$convertedData = $data[0];
+		}
+
+		return json_decode( $convertedData, true );
```
brucepearson: @strategio How about:
```php
return json_decode( is_array( $data ) ? $data[0] : $data, true );
```

src/Hooks/GutenbergCleanup.php
dgwatkins: ok 🤷

dgwatkins (dismissed):
Looks good

src/Hooks/GutenbergCleanup.php
strategio: Thanks @brucepearson (I will keep the closure variable, it make the logic more meaningfull to me).

src/DataConvert.php
strategio: I just extracted existing code @brucepearson. But OK, let's make it simple.

src/Hooks/GutenbergCleanup.php
```diff
@@ -0,0 +1,108 @@
...
+		if (
+			WPML_Elementor_Data_Settings::META_KEY_DATA === $metaKey
+			&& WPML_Elementor_Data_Settings::is_edited_with_elementor( $postId )
+		) {
+			$data = DataConvert::unserialize( $metaValue );
+
+			// $ifDataHasChanged :: array -> bool
+			$ifDataHasChanged = function( $processedData ) use ( $data ) {
+				return json_encode( $processedData ) !== json_encode( $data );
+			};
```
brucepearson: @strategio I would change this to make it more declarative where it's used.

```php
// $isDifferentFrom :: array -> array -> bool
$isDifferentFrom = curryN( 2, function( $data,  $processedData )  {
	return json_encode( $processedData ) !== json_encode( $data );
} );

...
...
->filter( $isDifferentFrom( $data ) )
```

strategio: Done @brucepearson.

src/Hooks/GutenbergCleanup.php
```diff
@@ -0,0 +1,109 @@
...
+				[ self::class, 'deletePackage' ]
+			) );
+
+			return Maybe::of( $data )
+				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
+				->filter( $ifDifferentFrom( $data ) )
+				->map( $updatePostMeta )
+				->map( $deleteGbPackage )
+				->map( Fns::always( true ) )
+				->getOrElse( $check );
```
brucepearson, line 58: @strategio Something doesn't seem right with this. I understand why you have done it this way. I would try to make it something like:

```php
return Maybe::of( $data )
				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
				->filter( $ifDifferentFrom( $data ) )
				->map( [ DataConvert::class, 'serialize' ] )
				->map( $updatePostMeta( $postId ) )
				->map( $deleteGbPackage )
				->getOrElse( $check );
```
and change $updatePostMeta to return $postId and $deleteGbPackage to return true.

This way the data flows through the functions.

Also $updatePostMeta and $deleteGbPackage are then pure functions and don't take data from the surrounding environment.

src/Hooks/GutenbergCleanup.php
brucepearson: I agree with @dgwatkins here. It doesn't hurt but adds confusion

src/Hooks/GutenbergCleanup.php
strategio: Please check again @brucepearson.

src/Hooks/GutenbergCleanup.php
strategio: Removed!

src/Hooks/GutenbergCleanup.php
brucepearson: @strategio I was thinking more like this. This way the data is passed from function to function and each function returns transformed data for the next function.

```php
			// $updatePostMeta :: int -> array -> int
			$updatePostMeta = curryN( 2, function( $postId, $meta ) {
				Post::updateMeta( $postId, WPML_Elementor_Data_Settings::META_KEY_DATA, $meta );
				return $postId;
			} );

			return Maybe::of( $data )
				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
				->filter( $ifDifferentFrom( $data ) )
				->map( [ DataConvert::class, 'serialize' ] )
				->map( $updatePostMeta( $postId ) )
				->map( [ self::class, 'getGutenbergPackage' ] )
				->map( [ self::class, 'deletePackage' ] )
				->getOrElse( $check );

...
...

public static function deletePackage( $package ) {
	$package && do_action( 'wpml_delete_package', $package->name, $package->kind );
	return true;    
}
```

The other option would be to have all the `side effects` in one function:

```php
			// $update :: int -> array -> bool
			$update = curryN( 2, function( $postId, $meta ) {
				Post::updateMeta( $postId, WPML_Elementor_Data_Settings::META_KEY_DATA, $meta );
				self::deletePackage( self::getGutenbergPackage( $postId ) );
				return true;
			} );

			return Maybe::of( $data )
				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
				->filter( $ifDifferentFrom( $data ) )
				->map( [ DataConvert::class, 'serialize' ] )
				->map( $update( $postId ) )
				->getOrElse( $check );

```

strategio: @brucepearson, I opted for the 2nd solution. Please check again!

src/Hooks/GutenbergCleanup.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @param string    $metaValue
+	 *
+	 * @return mixed
+	 */
+	public function removeGutenbergFootprint( $check, $postId, $metaKey, $metaValue ) {
+		if (
+			WPML_Elementor_Data_Settings::META_KEY_DATA === $metaKey
+			&& WPML_Elementor_Data_Settings::is_edited_with_elementor( $postId )
+		) {
+			$data = DataConvert::unserialize( $metaValue );
```
brucepearson: @strategio I think you can now move this into the Maybe:

```php
			return Maybe::of( $metaValue )
				->map( [ DataConvert::class, 'unserialize' ] )
				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
				->filter( $ifDifferentFrom( $data ) )
				->map( [ DataConvert::class, 'serialize' ] )
				->map( $update( $postId ) )
				->getOrElse( $check );
```

src/Hooks/GutenbergCleanup.php
brucepearson: @strategio Looks good now. Don't forget to delete commented out code!

brucepearson (dismissed):
@strategio Looks very nice. +1

src/Hooks/GutenbergCleanup.php
strategio: Unfortunately not @brucepearson, we need `$data` for the `$ifDifferentFrom` function. Or perhaps you have something to suggest?

src/Hooks/GutenbergCleanup.php
strategio: Ooops, good catch @brucepearson, thanks :)

src/Hooks/GutenbergCleanup.php
brucepearson: @strategio I guess you can't because we need `$data` in the filter function

brucepearson: @strategio Actually, maybe you could do:

```php
return Maybe::of( $metaValue )
				->map( [ DataConvert::class, 'unserialize' ] )
				->map( [ $this, 'removeBlockMetaInEditorWidget' ] )
				->map( [ DataConvert::class, 'serialize' ] )
				->filter( pipe( Relation::equals( $metaValue ), Fns::not() ) )
				->map( $update( $postId ) )
				->getOrElse( $check );
```

strategio: That's completely true @brucepearson, why would we make it complicated when it could be so simple :)

brucepearson (dismissed):

PR #122, opened 2020-09-04
[wpmlcore-7433] Fix fatal error for not loaded integration classes

GeorgeBotsev:
Add checking if we can find the integration class before instantiating it

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7433

### DIFF
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -53,20 +53,10 @@ class WPML_Elementor_Translatable_Nodes implements IWPML_Page_Builders_Translata
 					}
 				}
 
-				if ( isset( $node_data['integration-class'] ) ) {
-					foreach ( $this->get_integration_classes( $node_data ) as $class ) {
-						try {
-							if ( $class instanceof \WPML_Elementor_Module_With_Items ) {
-								$node = $class;
-							} elseif ( class_exists( $class ) ) {
-								$node = new $class();
-							}
-							if ( ! empty( $node ) ) {
-								$strings = $node->get( $node_id, $element, $strings );
-							}
-						} catch ( Exception $e ) {
-						}
-					}
+				foreach ( $this->get_integration_instances( $node_data ) as $instance ) {
+					try {
+						$strings = $instance->get( $node_id, $element, $strings );
+					} catch ( Exception $e ) {}
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


dgwatkins (dismissed):
lgtm @GeorgeBotsev

src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -59,9 +59,13 @@ public function get( $node_id, $element ) {
 							if ( $class instanceof \WPML_Elementor_Module_With_Items ) {
 								$node = $class;
 							} else {
-								$node = new $class();
+								if ( class_exists( $class ) ) {
```
strategio: @GeorgeBotsev, we can merge this into `elseif`:

```php
} elseif ( class_exists( $class ) ) {
    $node = new $class();
}
```

strategio (changes requested):
@GeorgeBotsev, this is the "get" part but we also need to do the same in the "update" part. See https://github.com/OnTheGoSystems/wpml-page-builders-elementor/blob/66e251fdd132941028536278e761a3049626e5c2/src/class-wpml-elementor-translatable-nodes.php#L112.

src/class-wpml-elementor-translatable-nodes.php
GeorgeBotsev: I noticed that we have the instantiation of the class at the other place too, but I did not get any error there when I tested.

strategio: @GeorgeBotsev, this is a good example of duplicated code where we have to apply the same fix in 2 different places. What we should do here is to refactor a little bit so we just have to fix in 1 place.

My suggestion is to replace the `get_integration_classes` method by `get_integration_instances` which will return a filtered array (only object instances).

strategio: Here's the idea @GeorgeBotsev https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commit/eacae8aa182b5bd981598ae60d8a9f63a125295d.

This is just the re-factoring and I think it could be a separate commit. Then you can add a second commit to specifically fix the issue (in 1 unique place this time).

It's on the branch https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commits/wpmlcore-7433-pierre.

Update - I had some failing tests and I had to revert some simplifications:

1. I had to keep the `try/catch` when we use the instances (https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commit/eacae8aa182b5bd981598ae60d8a9f63a125295d#diff-1e5046e82faeba27c319819ef7a502caR57) because it was making one test to fail. I think it's initially a problem in the test, but it also makes sense to keep it since integration classes are publicly opened.

2. `$node_data['integration-class']` can also contain a single instance (it was one of the test cases), so we cannot cast it to an array => https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commit/eacae8aa182b5bd981598ae60d8a9f63a125295d#diff-1e5046e82faeba27c319819ef7a502caR144.

GeorgeBotsev: Ok, I tested your suggestion and I applied the fix to check if the class exist, and I am implementing it in the code as a commit now.

strategio: @GeorgeBotsev, I suggest to separate the refactoring and the actual fix in 2 separate commits:
1. Re-factor duplicated code 
2. Prevent fatal error if integration class does not exist

This is a good pratice because:
1. We can make sure first that the re-factoring is not breaking the existing tests (the refactoring commit should not include any change in the tests).
2. It's easier to see the actual fix in the review (we can display only the commit related to the fix).

Moreover, I think you can also add a small unit test for the new case that you fixed (just pass a class name that does not exist, `An_Unexisting_Class`). This is not critical, but a good exercise for you to understand how we can progressively increase the constraints on our code (we don't change our tests, we just add more).

dgwatkins (dismissed):

strategio (dismissed):
+1 @GeorgeBotsev

PR #123, opened 2020-09-08
Remove data post meta from post hash when auto-update is activated

strategio:
**Goes with https://github.com/OnTheGoSystems/wpml-page-builders/pull/109**

We'll keep the old behavior in case the translation auto-update feature
is disabled.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7544

### DIFF
--- a/src/class-wpml-elementor-data-settings.php
+++ b/src/class-wpml-elementor-data-settings.php
@@ -124,7 +124,12 @@ class WPML_Elementor_Data_Settings implements IWPML_Page_Builders_Data_Settings 
 	 * @return array
 	 */
 	public function add_data_custom_field_to_md5( array $custom_fields_values, $post_id ) {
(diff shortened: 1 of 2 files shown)


src/class-wpml-elementor-data-settings.php
```diff
@@ -124,7 +124,12 @@ public function get_fields_to_save() {
 	 * @return array
 	 */
 	public function add_data_custom_field_to_md5( array $custom_fields_values, $post_id ) {
-		$custom_fields_values[] = get_post_meta( $post_id, $this->get_meta_field(), true );
+		if ( apply_filters( 'wpml_pb_is_auto_update_activated', false ) ) {
```
andreasciamanna: @strategio as asked in the other PR, can't you just check for `defined( 'WPML_TM_VERSION' )`?

PR #129, opened 2020-09-24
Translate section background image

GeorgeBotsev:
Call Media Translation function to translate the image with replacement image file (if set).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6506

### DIFF
--- a/src/media/class-wpml-elementor-media-node-provider.php
+++ b/src/media/class-wpml-elementor-media-node-provider.php
@@ -56,6 +56,10 @@ class WPML_Elementor_Media_Node_Provider {
 					$node = new WPML_Elementor_Media_Node_WP_Widget_Media_Gallery( $this->media_translate );
 					break;
 
+				case 'all_nodes':
+					$node = new \WPML\PB\Elementor\Media\Modules\AllNodes( $this->media_translate );
+					break;
+
 				default:
 					$node = null;
 			}
--- a/src/media/class-wpml-elementor-media-nodes-iterator.php
+++ b/src/media/class-wpml-elementor-media-nodes-iterator.php
@@ -23,6 +23,10 @@ class WPML_Elementor_Media_Nodes_Iterator implements IWPML_PB_Media_Nodes_Iterat
 			} elseif ( $this->is_valid_media_node( $node ) ) {
 				$node = $this->translate_node( $node, $lang, $source_lang );
 			}
+
+			if ( isset( $node['settings'] ) ) {
+				$node["settings"] = $this->node_provider->get( 'all_nodes' )->translate( $node['settings'], $lang, $source_lang );
+			}
 		}
 
 		return $data_array;
--- a/src/media/modules/AllNodes.php
+++ b/src/media/modules/AllNodes.php
@@ -1,0 +1,31 @@
+<?php
+
+
+namespace WPML\PB\Elementor\Media\Modules;
+
+
+class AllNodes extends \WPML_Elementor_Media_Node {
+
+	/**
+	 * @param array $settings
+	 * @param string $target_lang
+	 * @param string $source_lang
+	 *
+	 * @return array
+	 */
+	public function translate( $settings, $target_lang, $source_lang ) {
+		$backgrounds = [
+			'_background_image',
+			'background_image',
+			'background_overlay_image',
+			'background_hover_image'
+		];
+
+		foreach ( $backgrounds as $background ) {
+				$settings = $this->translate_image_property( $settings, $background , $target_lang,
+					$source_lang );
+		}
+
+		return $settings;
+	}
+}
--- a/tests/phpunit/tests/media/modules/test-AllNodes.php
+++ b/tests/phpunit/tests/media/modules/test-AllNodes.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace WPML\PB\Elementor\Media\Modules;
+
+use WPML\FP\Math;
+
+/**
+ * @group media
+ */
+class Test_AllNodes extends \OTGS_TestCase {
+
+	/**
+	 * @test
+	 */
+	public function it_translates_background_images() {
+		$settings = [
+			'structure'              => 20,
+			'background_image'       => [
+				'url' => 'background_image_1',
+				'id'  => 1,
+			],
+			'background_size'        => 'contain',
+			'background_hover_image' => [
+				'url' => 'background_image_2',
+				'id'  => 2,
+			],
+		];
+
+		$source = 'en';
+		$target = 'fr';
+
+		$expected = [
+			'structure'              => 20,
+			'background_image'       => [
+				'url' => 'translated_background_image_1',
+				'id'  => 101,
+			],
+			'background_size'        => 'contain',
+			'background_hover_image' => [
+				'url' => 'translated_background_image_2',
+				'id'  => 102,
+			],
+		];
+
+		$mediaTranslate = $this->createMock( '\WPML_Page_Builders_Media_Translate' );
+		$mediaTranslate->method( 'translate_id' )->willReturnCallback( Math::add( 100 ) );
+		$mediaTranslate->method( 'translate_image_url' )->willReturnMap( [
+			[ 'background_image_1', $target, $source, 'translated_background_image_1' ],
+			[ 'background_image_2', $target, $source, 'translated_background_image_2' ],
+		] );
+
+		$subject = new AllNodes( $mediaTranslate );
+
+		$this->assertEquals( $expected, $subject->translate( $settings, $target, $source ) );
+	}
+
+}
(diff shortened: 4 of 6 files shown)


src/media/class-wpml-elementor-media-nodes-iterator.php
```diff
@@ -19,6 +19,9 @@ public function __construct( WPML_Elementor_Media_Node_Provider $node_provider )
 	public function translate( $data_array, $lang, $source_lang ) {
 		foreach ( $data_array as &$node ) {
 			if ( $this->is_parent_node( $node ) ) {
+				if ( $this->is_background_image_set( $node ) ) {
+					$node = $this->translate_node_background_image( $node, $source_lang, $lang );
```
strategio: @GeorgeBotsev, when we started this project we were only translating strings (in this class).

Then, we started supporting media translation in a separate layer. See https://github.com/OnTheGoSystems/wpml-page-builders-elementor/blob/master/src/media/class-wpml-elementor-media-nodes-iterator.php#L3.

Check if you can insert your code in `WPML_Elementor_Media_Nodes_Iterator` or maybe in the base module class `WPML_Elementor_Media_Node`.

strategio: If I understand the problem correctly, we should convert the background image on any king of element (widget or node, not sure how we should call it). So maybe we need a generic module extending `WPML_Elementor_Media_Node` (e.g. `WPML_Elementor_All_Widget_Media_Node`) that we could get from the `WPML_Elementor_Media_Node_Provider`.

Then in `WPML_Elementor_Media_Nodes_Iterator`:

```php
	public function translate( $data_array, $lang, $source_lang ) {
		foreach ( $data_array as &$node ) {
			if ( $this->is_parent_node( $node ) ) {
				$node['elements'] = $this->translate( $node['elements'], $lang, $source_lang );
			} elseif ( $this->is_valid_media_node( $node ) ) {
				$node = $this->translate_node( $node, $lang, $source_lang );
			}

			// => Here we'll use this "all widget" media node - not sure the naming is correct though
			$node = $this->node_provider->get( 'all_widgets' )->translate( $node_data['settings'], $lang, $source_lang );
		}

		return $data_array;
	}
```

GeorgeBotsev: > Check if you can insert your code in `WPML_Elementor_Media_Nodes_Iterator`

I am a bit puzzled.
My code is in  `WPML_Elementor_Media_Nodes_Iterator` class.

GeorgeBotsev: I changed the code - moved it in a separate class.
Let me know what you think @strategio

src/modules/AllWidgets.php
```diff
@@ -0,0 +1,30 @@
...
+	 * @param string $source_lang
+	 *
+	 * @return array
+	 */
+	public function translate( $settings, $target_lang, $source_lang ) {
+		global $sitepress;
+		$background = [ 'background_image', 'background_overlay_image', 'background_hover_image'] ;
+		foreach ( $background as $element ) {
+			if ( isset( $settings[$element]["url"] ) ) {
+				$media_translate                     = new \WPML_Media_Image_Translate( $sitepress,
```
strategio: @GeorgeBotsev, you don't need to instantiate this object because you already have it as a dependency:

```php
$settings[$element]["url"] = $this->media_translate->get_translated_image_by_url( $settings[$element]["url"], $source_lang, $target_lang );
```

And actually we already have a better parent protected method for this so you can simplify with:

```php
$backgrounds = [
    'background_image',
    'background_overlay_image',
    'background_hover_image'
] ;

foreach ( $backgrounds as $background ) {
    $settings = $this->media_translate->translate_image_property( $settings, $background, $target_lang, $source_lang );
}

return $settings;
```

GeorgeBotsev:
I found that if a text block for an example, has a background, it is saved in a key "_background_image".
We have logic in \WPML_Elementor_Media_Node_Image::translate but it is not being run because of the node type, when the page is translated.

src/modules/AllWidgets.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+
+namespace WPML\PB\Elementor\Modules;
```
strategio: @GeorgeBotsev, we have a dedicated folder for media stuff: `src/media/modules` and so the namespace should be `WPML\PB\Elementor\Media\Modules`.

src/modules/AllWidgets.php
```diff
@@ -0,0 +1,30 @@
...
+		foreach ( $background as $element ) {
+			if ( isset( $settings[ $element ]["url"] ) ) {
+				$media_translate             = new \WPML_Media_Image_Translate( $sitepress,
+					new \WPML_Media_Attachment_By_URL_Factory() );
+				$settings[ $element ]["url"] = $media_translate->get_translated_image_by_url( $settings[ $element ]["url"],
+					$source_lang, $target_lang );
+			}
+		}
+
+		return $settings;
```
strategio: You can simplify the whole method:

```php
$backgrounds = [
    'background_image',
    'background_overlay_image',
    'background_hover_image'
];

foreach ( $backgrounds as $background ) {
	$settings = $this->media_translate->translate_image_property( $settings, $background, $target_lang, $source_lang );
}

return $settings;
```

You don't need to instantiate `WPML_Media_Image_Translate` because it's already a dependency of this class.

Generally, we should instantiate classes only in factories. Otherwise it does not allow proper testing because we can't mock the object.

strategio (changes requested):
Nice @GeorgeBotsev, this is what I was expecting.

Sorry if you already implemented your code in the `WPML_Elementor_Media_Nodes_Iterator`, probably I got confused.

GeorgeBotsev:
I found a problem with my code. If the widget/block/whatever element is image, then the logic is being run and the image is being returned as "false" and as a result - no image shown on the translation.

src/media/class-wpml-elementor-media-node-provider.php
```diff
@@ -56,6 +56,10 @@ public function get( $type ) {
 					$node = new WPML_Elementor_Media_Node_WP_Widget_Media_Gallery( $this->media_translate );
 					break;
 
+				case 'all_widgets':
+					$node = new \WPML\PB\Elementor\Modules\AllWidgets( $this->media_translate );
```
strategio: @GeorgeBotsev, maybe it would be more accurate to call this `AllNodes`. I am not sure about the terminology but I understand that this applies to all "nodes" and I am not sure that all "nodes" are "widgets".

src/modules/AllWidgets.php
```diff
@@ -0,0 +1,30 @@
...
+class AllWidgets extends \WPML_Elementor_Media_Node {
+
+	/**
+	 * @param array $settings
+	 * @param string $target_lang
+	 * @param string $source_lang
+	 *
+	 * @return array
+	 */
+	public function translate( $settings, $target_lang, $source_lang ) {
```
strategio: Please use camelCase when possible for this new class.

src/modules/AllWidgets.php
GeorgeBotsev: I addressed that now.

src/modules/AllWidgets.php
GeorgeBotsev: fixed.

src/media/class-wpml-elementor-media-node-provider.php
GeorgeBotsev: ok, I renamed the file and case.

src/modules/AllWidgets.php
GeorgeBotsev: The class is now moved there as requested.

src/media/modules/allNodes.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+
+namespace WPML\PB\Elementor\Media\Modules;
+
+
+class allNodes extends \WPML_Elementor_Media_Node {
```
strategio: @GeorgeBotsev, the class name should start with uppercase `AllNodes`.

strategio (commented):
Looks good to me @GeorgeBotsev!

If you confirm this last version is working, then I can take over to add unit tests.

src/media/modules/allNodes.php
```diff
@@ -0,0 +1,33 @@
...
+	public function translate( $settings, $target_lang, $source_lang ) {
+		$backgrounds = [
+			'_background_image',
+			'background_image',
+			'background_overlay_image',
+			'background_hover_image'
+		];
+
+		foreach ( $backgrounds as $background ) {
+			if ( isset ( $settings[ $background ]['url'] ) ) {
```
strategio: I think this check is not needed because it's already done inside `translate_images_property`.

src/media/modules/allNodes.php
GeorgeBotsev: OK, fixing this now.

src/media/modules/allNodes.php
GeorgeBotsev: In my previous code - it had to be there, but now my new code is not doing its job and I am checking on why.

GeorgeBotsev: It is fixed now, the line is removed.

gordonsh24:
@OnTheGoSystems/compatibility @BruceOnTheGo Guys, George's fix seems to work smoothly. I covered it with unit tests. Please review it.

gregWierzba (dismissed):

dgwatkins (dismissed):
:+1:

PR #130, opened 2020-09-30
Fix Elementor posts widget pagination

mrudzki91:
Fix Elementor posts widget pagination when page is displayed as translated

It uses the filter from MR: https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/-/merge_requests/2884

wpmlcore-7482

### DIFF
--- a/src/class-wpml-elementor-urls.php
+++ b/src/class-wpml-elementor-urls.php
@@ -23,6 +23,7 @@ class WPML_Elementor_URLs implements IWPML_Action {
 
 	public function add_hooks() {
 		add_filter( 'elementor/document/urls/edit', array( $this, 'adjust_edit_with_elementor_url' ), 10, 2 );
+		add_filter( 'wpml_is_pagination_url_in_post', [ $this, 'is_pagination_url' ], 10, 3 );
 	}
 
 	public function adjust_edit_with_elementor_url( $url, $elementor_document ) {
@@ -36,5 +37,22 @@ class WPML_Elementor_URLs implements IWPML_Action {
 		}
 
 		return $this->language_converter->convert_admin_url_string( $url, $post_language );
+	}
+
+	/**
+	 * Check if the given URL is the pagination inside the post.
+	 *
+	 * @param bool $is_pagination_url_in_post
+	 * @param string $url URL to check
+	 * @param string $post_name Current post name
+	 *
+	 * @return bool
+	 */
+	public function is_pagination_url( $is_pagination_url_in_post, $url, $post_name ) {
+		return $is_pagination_url_in_post
+		       || (
+			       WPML_Elementor_Data_Settings::is_edited_with_elementor( get_the_ID() )
+			       && preg_match_all( "/$post_name\/([\d]*)\/$/", $url )
+		       );
 	}
 }
--- a/src/class-wpml-elementor-urls.php
+++ b/src/class-wpml-elementor-urls.php
@@ -24,6 +24,7 @@ class WPML_Elementor_URLs implements IWPML_Action {
 	public function add_hooks() {
 		add_filter( 'elementor/document/urls/edit', array( $this, 'adjust_edit_with_elementor_url' ), 10, 2 );
 		add_filter( 'wpml_is_pagination_url_in_post', [ $this, 'is_pagination_url' ], 10, 3 );
+		add_filter( 'paginate_links', [ $this, 'fix_pagination_link_with_language_param' ], 10, 1 );
 	}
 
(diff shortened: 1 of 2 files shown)


src/class-wpml-elementor-urls.php
```diff
@@ -23,6 +23,7 @@ public function __construct(
 
 	public function add_hooks() {
 		add_filter( 'elementor/document/urls/edit', array( $this, 'adjust_edit_with_elementor_url' ), 10, 2 );
+		add_filter( 'wpml_is_pagination_in_post', [ $this, 'is_pagination_in_post' ], 10, 3 );
```
andreasciamanna: @mrudzki91 the filter's handle name is a bit confusing, in my opinion.

I had to look at the code and the implementation in the other MR to understand what it's supposed to do.

Can you please give it a better name?

The information I think are missing in the handle name are:

- Q: Which post are we talking about? A: The current one.
- Why pagination? This is term is usually used for the handling of the pagination (business logic), or for UI elements that handles the pagination. 
   Here instead we want to know if the (current) post is "paginated". 
   In fact, is not just the post, but the current URL/path of the current post (I believe because I'm still not that sure I understand how it works).

Perhaps something like `wpml_is_current_post_paginated`, even if it's still missing the information about the "URL", but adding that as well, would make the handle a bit too long.

andreasciamanna (changes requested):

src/class-wpml-elementor-urls.php
mrudzki91: @andreasciamanna I've renamed the filter to `wpml_is_pagination_url_in_post` and also changed some variable names to make sure it's more clear.
This filter checks if the given URL is a pagination URL inside the given post (so it doesn't check the current post URL but some other URL inside the post `$post_name`).

andreasciamanna (dismissed):
Thanks. It's better now!

src/class-wpml-elementor-urls.php
```diff
@@ -37,4 +38,21 @@ public function adjust_edit_with_elementor_url( $url, $elementor_document ) {
...
+	 * @param bool $is_pagination_url_in_post
+	 * @param string $url URL to check
+	 * @param string $post_name Current post name
+	 *
+	 * @return bool
+	 */
+	public function is_pagination_url( $is_pagination_url_in_post, $url, $post_name ) {
+		return $is_pagination_url_in_post
+		       || (
+			       get_post_meta( get_the_ID(), '_elementor_edit_mode', true )
```
strategio: @mrudzki91, we have a class constant `WPML_Elementor_Data_Settings::META_KEY_MODE` for this.

And actually, I think you should use this instead:

```php
WPML_Elementor_Data_Settings::is_edited_with_elementor( get_the_ID() )
```

It checks if the post meta is equal to `builder` but I think this is what we actually want in the end.

mrudzki91: @strategio thanks, I've replaced it. ;)

strategio: Thanks @mrudzki91 :)

strategio (dismissed):

mrudzki91:
I've added a fix for Elementor pagination URLs when using Language as a parameter (URLs from Elementor are like `http://example.com/post_name/?lang=pl/2/`)
Please, take another look.

src/class-wpml-elementor-urls.php
```diff
@@ -37,4 +39,40 @@ public function adjust_edit_with_elementor_url( $url, $elementor_document ) {
...
+	 * @param string $post_name Current post name
+	 *
+	 * @return bool
+	 */
+	public function is_pagination_url( $is_pagination_url_in_post, $url, $post_name ) {
+		return $is_pagination_url_in_post
+		       || (
+			       WPML_Elementor_Data_Settings::is_edited_with_elementor( get_the_ID() )
+			       && (
+				       preg_match_all( "/$post_name\/([\d]*)\/$/", $url )
```
brucepearson: @mrudzki91 Should this be {$post_name} in the regex string?

mrudzki91: @brucepearson , sure, changed.

brucepearson: @mrudzki91 I think this shows that the unit test is not checking for the case of returning false. You probably need a unit test that tests the case of the regex not matching.

brucepearson: @mrudzki91 You can ignore my comments. I always thought you had to have the curly brackets.

brucepearson (dismissed):
@mrudzki91 +1

PR #131, opened 2020-10-01
[wpmlcore-7615] Fix Media Carousel Video URL not translatable

izzi-hassan:
I tried to provide individual titles for the `image_link_to` and the `video` URL fields but the changes and tests required are probably not worth it since we are moving towards XML config for these widgets.

### DIFF
--- a/src/modules/MediaCarousel.php
+++ b/src/modules/MediaCarousel.php
@@ -15,7 +15,10 @@ class MediaCarousel extends \WPML_Elementor_Module_With_Items {
 	 * @return array
 	 */
 	public function get_fields() {
-		return [ 'image_link_to' => [ 'url' ] ];
+		return [ 
+			'image_link_to' => [ 'url' ],
+			'video' => [ 'url ']
+		];
 	}
 
 	/**
@@ -26,7 +29,7 @@ class MediaCarousel extends \WPML_Elementor_Module_With_Items {
 	protected function get_title( $field ) {
 		switch ( $field ) {
 			case 'url':
-				return esc_html__( 'Media Carousel: link URL', 'sitepress' );
+				return esc_html__( 'Media Carousel: slide URL', 'sitepress' );
 			default:
 				return '';
 		}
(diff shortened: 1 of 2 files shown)


src/modules/MediaCarousel.php
```diff
@@ -15,7 +15,10 @@ public function get_items_field() {
 	 * @return array
 	 */
 	public function get_fields() {
-		return [ 'image_link_to' => [ 'url' ] ];
+		return [
```
amElnagdy, line 18: did it work without field_id?

izzi-hassan: Out of the box, you can't use `field` and `field_id` with the compatibility class.

This REVERTED commit shows what I needed to do to use `field_id`: 
https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commit/87e6779dfa2d9694aca69d0e305dcc8b066da23e

^I am not sure how this would affect other widgets.

amElnagdy: Well, if it is going to affect other widgets, I think this can wait for the new wpml-config.xml file.

izzi-hassan: I think there's a bit of confusion here. I REVERTED the commit that would affect other widgets.

This MR simply registers the video field in MediaCarousel.php and adjusts the Test class accordingly.

amElnagdy: 😅  yes, sorry it's my fault.

izzi-hassan: Thanks, I CAN be confusing 😛

amElnagdy (approved):
👍

strategio:
@amElnagdy @izzi-hassan, I suggest that we wait for wpmlcore-7241 (Move the API page builders configuration to the `wpml-config.xml` file) to fix it in the WPML config file instead.

@amElnagdy already started trying to transpose everything, but we can also do this in an incremental way: The XML config will overwrite the PHP one.

What do you think?

amElnagdy:
@strategio, it can wait yes.

strategio:
@izzi-hassan, now that we have the [XML config available for API page builders](https://wpml.org/documentation/support/language-configuration-files/how-to-register-page-builder-widgets-for-translation/), do you think you could transpose this widget config into XML? And then open a MR in https://github.com/OnTheGoSystems/wpml-config instead?

izzi-hassan:
@strategio absolutely! I'll submit an MR there in the `wpml-config` repo.

strategio:
Thanks @izzi-hassan, closing here.

PR #133, opened 2020-10-21
Add path support for fields

strategio:
**This is an implementation specific to Elementor, as we have this case only here.**

We can now easily translate this data shape:

```php
[
    'widgetType' => 'my-widget',
    'settings' => [
        'foo' => [
            'bar' => 'The string to translate',
        ],
    ],
]
```

With the following config:

```xml
<wpml-config>
    <elementor-widget>
        <widget name="my-widget">
            <fields>
                <field type="The foo bar thing">foo>bar</field>
            </fields>
        </widget>
    </elementor-widget>
</wpml-config>
```
~~The `|` separator is arbitrary, let me know if you have a better suggestion.~~ As Bruce suggested, I used `>` as a separator.

FTR, we used to have associative keys for such cases and this is making the code generally more complex (AFAIK, we used this only for URL fields). Our initial strategy was probably not the best, but it was hard to anticipate everything.

I managed to do this and keep the backward compatibility. This will be way more flexible for us.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7676

### DIFF
--- a/src/class-wpml-elementor-translatable-nodes.php
+++ b/src/class-wpml-elementor-translatable-nodes.php
@@ -1,5 +1,6 @@
 <?php
 
+use WPML\FP\Obj;
 use WPML\PB\Elementor\DynamicContent\Strings as DynamicContentStrings;
 use WPML\PB\Elementor\Modules\ModuleWithItemsFromConfig;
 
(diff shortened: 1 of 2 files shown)


src/class-wpml-elementor-translatable-nodes.php
```diff
@@ -108,27 +114,12 @@ public function update( $node_id, $element, WPML_PB_String $string ) {
...
-	 * @param string     $field_key
-	 *
-	 * @return bool
+	 * @return string[]
 	 */
-	private function is_array_field( $element, $field_wrapper, $field_key ) {
-		return isset( $element[ self::SETTINGS_FIELD ][ $field_wrapper ][ $field_key ] )
-		       && trim( $element[ self::SETTINGS_FIELD ][ $field_wrapper ][ $field_key ] );
+	private static function get_partial_path( $field ) {
+		return explode( '|', $field );
```
brucepearson: @strategio Would `>` work? eg `foo>bar`? I'm not sure if this would cause an issue in wpml_config.xml.

brucepearson (dismissed):
@strategio It looks ok.

src/class-wpml-elementor-translatable-nodes.php
strategio: Yes @brucepearson it works with `>`. I tested this in WPML > Settings > Custom XML Config, no problem. And string registration, data translation works perfectly. I changed it!

strategio:
@OnTheGoSystems/compatibility, please check this.

dgwatkins (dismissed):
Looks good

PR #140, opened 2021-01-13
Cast to string if what we received was a string. 

dgwatkins:
Fixes issues with  strict  comparisons in elementor.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7878

### DIFF
--- a/tests/phpunit/tests/test-wpml-elementor-translate-ids.php
+++ b/tests/phpunit/tests/test-wpml-elementor-translate-ids.php
@@ -32,12 +32,9 @@ class Test_WPML_Elementor_Translate_IDs extends OTGS_TestCase {
 
 	/**
 	 * @test
+	 * @dataProvider dp_translate_template_id
 	 */
-	public function it_should_translate_template_id() {
-
-		$template_id   = 123;
-		$translated_id = 456;
-		$post_type     = 'anything';
+	public function it_should_translate_template_id( $template_id, $translated_id, $post_type ) {
 
 		\WP_Mock::userFunction( 'get_post_type',
 			array(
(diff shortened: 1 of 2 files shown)


strategio (commented):
@dgwatkins, it looks good to me but I think it should not be hard to cover it with a unit test.

You can reuse `Test_WPML_Elementor_Translate_IDs::it_should_translate_template_id` with a data provider.

dgwatkins:
> @dgwatkins, it looks good to me but I think it should not be hard to cover it with a unit test.
> 
> You can reuse `Test_WPML_Elementor_Translate_IDs::it_should_translate_template_id` with a data provider.

I'm working on it, that's why I didn't call for review yet :-) thanks for the notice anyway

dgwatkins:
@strategio I used your suggestion, thanks!

@OnTheGoSystems/compatibility can I have two reviews please?

strategio (dismissed):
+1 @dgwatkins

tests/phpunit/tests/test-wpml-elementor-translate-ids.php
```diff
@@ -48,14 +45,21 @@ public function it_should_translate_template_id() {
...
 
 		$this->assertEquals( $translated_id, $subject->translate_theme_location_template_id( $template_id ) );
 
 	}
 
+	public function dp_translate_template_id() {
+		return [
+			[ 123, 456, 'anything' ],
+			[ '123', '456', 'anything' ],
+		];
```
brucepearson, line 60: @dgwatkins I would add keys so it's clear why these two are different.
```php
return [
	'preserve integer type' => [ 123, 456, 'anything' ],
	'preserve string type'   => [ '123', '456', 'anything' ],
];
```

brucepearson (dismissed):
One minor comment. +1

tests/phpunit/tests/test-wpml-elementor-translate-ids.php
dgwatkins: I handled in https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/141 because this was already merged when I got here.

PR #142, opened 2021-01-15
Handle in_tax as well as in_tax_children template conditions

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7674

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,7 @@
   },
   "scripts": {
     "post-update-cmd": [
-      "sed -i 's/final public function add_group_control/public function add_group_control/g' ./vendor/elementor/elementor/includes/base/controls-stack.php"
+      "sed -i'' 's/final public function add_group_control/public function add_group_control/g' ./vendor/elementor/elementor/includes/base/controls-stack.php"
     ]
   }
 }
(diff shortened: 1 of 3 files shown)


dgwatkins:
review please :-)
@OnTheGoSystems/compatibility 
@OnTheGoSystems/wpml

strategio (dismissed):
+1 @dgwatkins

dgwatkins:
@strategio I need to re-request your review because the data provider was incorrect.
A small change to be able to build on both Linux & Mac also slipped in.

strategio (dismissed):

composer.json
```diff
@@ -54,7 +54,7 @@
   },
   "scripts": {
     "post-update-cmd": [
-      "sed -i 's/final public function add_group_control/public function add_group_control/g' ./vendor/elementor/elementor/includes/base/controls-stack.php"
+      "sed -i'' 's/final public function add_group_control/public function add_group_control/g' ./vendor/elementor/elementor/includes/base/controls-stack.php"
```
andreasciamanna, line 57: @dgwatkins can you explain this change?

dgwatkins: @andreasciamanna I didn't really mean to commit that change. It actually slipped in during a fixup commit.

However, the change is required to run unit tests on my local host (Mac)

Linux just takes `-i` with no arguments.
Mac takes `-i` with the extension of the bak file (ie: `.bak`).

This way of calling it `-i''` (with no spaces) plays well with both Mac & Linux versions of sed.

andreasciamanna: Good to know!

It would be better if you could move this change in a separate commit where you explain the reason of the change.

andreasciamanna:
Anyway, +1

andreasciamanna (dismissed):

composer.json
dgwatkins: I see you approved, so I will make sure I issue a separate commit next time.

PR #143, opened 2021-01-15
Translate product_id in wc-add-to-cart widget

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7714

### DIFF
--- a/src/class-wpml-elementor-translate-ids.php
+++ b/src/class-wpml-elementor-translate-ids.php
@@ -1,5 +1,7 @@
 <?php
 
+use WPML\FP\Obj;
+
 /**
  * Class WPML_Elementor_Translate_IDs
  */
@@ -25,6 +27,7 @@ class WPML_Elementor_Translate_IDs implements IWPML_Action {
 			'translate_template_id'
 		) );
 		add_filter( 'elementor/frontend/builder_content_data', array( $this, 'translate_global_widget_ids' ), 10, 2 );
+		add_filter( 'elementor/frontend/builder_content_data', array( $this, 'translate_product_ids' ), 10, 2 );
 	}
 
 	public function translate_theme_location_template_id( $template_id ) {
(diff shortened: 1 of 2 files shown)


dgwatkins:
Review please
@OnTheGoSystems/compatibility @OnTheGoSystems/wpml

kkarpieszuk:
+1

kkarpieszuk (dismissed):

src/class-wpml-elementor-translate-ids.php
```diff
@@ -103,6 +104,25 @@ public function translate_global_widget_ids( $data_array, $post_id ) {
...
 
+	/**
+	 * @param array $data_array
+	 * @param int   $post_id
+	 *
+	 * @return array
+	 */
+	public function translate_product_ids( $data_array, $post_id ) {
+		foreach ( $data_array as &$data ) {
+			if ( isset( $data['elType'] ) && 'widget' === $data['elType'] ) {
```
strategio: You can simplify with only 1 `if`:

```php
if ( Obj::prop( 'elType', $data ) === 'widget' && Obj::prop( 'widgetType', $data ) === 'wc-add-to-cart' ) {
    $data['settings']['product_id'] = $this->translate_id( $data['settings']['product_id'] );
}
```

If you want to go the extra mile, this is a common pattern to use functional programming with `over` and a `lensProp`.

Now that you have the unit tests you could even try to refactor your code with a full FP pipeline (I'll let you decide).

tests/phpunit/tests/test-wpml-elementor-translate-ids.php
```diff
@@ -399,4 +399,46 @@ public function it_should_translate_location_condition_any_child_of() {
...
+	 * @test
+	 * @group wpmlcore-7714
+	 */
+	public function it_translates_product_ids() {
+		$post_id       = 123;
+		$product_id    = 456;
+		$tr_product_id = 789;
+
+		$data = [ [
+			'id' => '4a3eaf7',
```
strategio: @dgwatkins, I would use a 'wc-add-to-cart' inside another widget so we can actually test the recursive part of the method, otherwise the test will pass even if we remove `$data['elements'] = $this->translate_product_ids( $data['elements'], $post_id );`.

strategio (changes requested):

tests/phpunit/tests/test-wpml-elementor-translate-ids.php
dgwatkins: @strategio added one more level. Now we have section > column > widget.

src/class-wpml-elementor-translate-ids.php
dgwatkins: I am giving up @strategio 

I need to be more familiar with the different patterns to be able to apply them.

I did apply your suggestion though

strategio (dismissed):
It's fine @dgwatkins, thanks for trying +1

(I will try to do this in a separate MR just to show you as an example)

dgwatkins:
> It's fine @dgwatkins, thanks for trying +1
> 
> (I will try to do this in a separate MR just to show you as an example)

That would be a great lesson

kkarpieszuk (dismissed):

PR #144, opened 2021-01-18
Translate product Ids in FP way

strategio:
@dgwatkins, here's an example to transpose your code to FP. It's far from perfect, we should find a way to get rid of the `if`... but the idea is here, mostly about `over` and `lens`.

### DIFF
--- a/src/class-wpml-elementor-translate-ids.php
+++ b/src/class-wpml-elementor-translate-ids.php
@@ -1,6 +1,7 @@
 <?php
 
 use WPML\FP\Obj;
+use function WPML\FP\partialRight;
 
 /**
  * Class WPML_Elementor_Translate_IDs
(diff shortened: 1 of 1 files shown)


src/class-wpml-elementor-translate-ids.php
```diff
@@ -113,15 +114,26 @@ public function translate_global_widget_ids( $data_array, $post_id ) {
...
 
-		return $data_array;
+		$applyRecursivelyOnElements = Obj::over(
+			Obj::lensProp( 'elements' ),
+			partialRight( [ $this, 'translate_product_ids' ], $post_id )
+		);
+
+		return wpml_collect( $data_array )
+			->map( $translateProductId )
+			->map( $applyRecursivelyOnElements )
```
strategio, line 135: We could also have only one `map` => `->map( pipe( $translateProductId, $applyRecursivelyOnElements ) )`

dgwatkins (approved):
@strategio thanks for that!

looks beautiful now

I will even bookmark this example

PR #145, opened 2021-04-02
Elementor PRO - Gallery widget: cannot translate the gallery images t…

kkarpieszuk:
…itle, caption, desc

- added class WPML_Elementor_Media_Node_Gallery for gallery

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7531


dgwatkins:
@kkarpieszuk 
changes look good! :+1:

any idea what's up with unit tests?

dgwatkins (dismissed):

src/media/modules/class-wpml-elementor-media-node-gallery.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+class WPML_Elementor_Media_Node_Gallery extends \WPML_Elementor_Media_Node {
```
strategio: @kkarpieszuk, I guess we already abstracted this data shape. Did you try this?

```php
namespace WPML\PB\Elementor\Media\Modules;
    
class Gallery extends \WPML_Elementor_Media_Node_With_Images_Property {
    protected function get_property_name() {
        return 'gallery';
    }
}
```

What we are doing here is probably the same as `\WPML_Elementor_Media_Node_Image_Gallery`.

(please use a namespace for the new classes)

strategio (commented):
@kkarpieszuk, could you also cover the changes with unit tests?

strategio:
I could not fix the problem with the Github Actions, we'll handle it in a separate ticket (https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-191).

kkarpieszuk:
@strategio / @dgwatkins modified according to your suggestions plus covered with tests. Please continue the review

src/media/modules/class-wpml-elementor-media-node-gallery.php
strategio: @kkarpieszuk, I can't see what I suggested in your latests commits.

Also, please rename the file `Gallery.php`.

kkarpieszuk: @strategio file renamed. 

But I didn't get what you are asking for here. To introduce `get_property_name()` method? I don't need it, all is needed here is `translate()` method which looks for `gallery` array element and for each found, translate each internal image array

kkarpieszuk: wait... I think I understand now: if I declare class as extending `WPML_Elementor_Media_Node_With_Images_Property` (I didn't notice it) all I might need is to define property name, got it. Let me try...

kkarpieszuk: @strategio nope, it doesn't work, seems like this kind of element does not have `WPML_Page_Builders_Media_Translate` structure same as the rest which are translated with `WPML_Elementor_Media_Node_With_Images_Property`

kkarpieszuk:
@dgwatkins @strategio I am closing this one and opening new one, clean with suggestions from Pierre applied
https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/146

I can't work with this branch anymore (I actually could but it would take much more time than creating new branch and MR, I already spent hours to figure out why it is not working for me, see https://chat.onthegosystems.com/otgs/pl/g4ywqmeeuffbjd1a3j33jhmfna )

Anyway, please go to new pull request, review (it is the same as here + all suggestions from Pierre applied and are working) and give thumbs up if it is ok

PR #147, opened 2021-05-13
Fix incorrect post URL for landing pages

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8369

### DIFF
--- a/src/Hooks/LandingPages.php
+++ b/src/Hooks/LandingPages.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace WPML\PB\Elementor\Hooks;
+
+use WPML\FP\Obj;
+
+class LandingPages implements \IWPML_Frontend_Action, \IWPML_Backend_Action, \IWPML_DIC_Action {
+
+	const POST_TYPE = 'e-landing-page';
+
+	/** @var \SitePress $sitepress */
+	private $sitepress;
+
+	public function __construct( \SitePress $sitepress ) {
+		$this->sitepress = $sitepress;
+	}
+
+	public function add_hooks() {
+		if ( get_option( 'permalink_structure' ) ) {
+			add_filter( 'post_type_link', [ $this, 'adjustLink' ], PHP_INT_MAX, 2 );
+		}
+	}
+
+	/**
+	 * @see \Elementor\Modules\LandingPages\Module::remove_post_type_slug
+	 *
+	 * @param string   $postUrl
+	 * @param \WP_Post $post
+	 *
+	 * @return string
+	 */
+	public function adjustLink( $postUrl, $post ) {
+		if ( self::POST_TYPE !== $post->post_type || 'publish' !== $post->post_status) {
+			return $postUrl;
+		}
+
+		$homeUrl          = get_home_url();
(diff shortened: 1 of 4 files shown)


strategio:
@OnTheGoSystems/compatibility, please check this!

src/Hooks/LandingPages.php
```diff
@@ -0,0 +1,43 @@
...
+	/**
+	 * @see \Elementor\Modules\LandingPages\Module::remove_post_type_slug
+	 *
+	 * @param string   $postUrl
+	 * @param \WP_Post $post
+	 *
+	 * @return string
+	 */
+	public function adjustLink( $postUrl, $post ) {
+		if ( self::POST_TYPE === $post->post_type ) {
```
dgwatkins: @strategio Wouldn't it better to use `get_post_type()`?

src/Hooks/LandingPages.php
```diff
@@ -0,0 +1,43 @@
...
+			$homeUrl          = get_home_url();
+			$urlParts         = wp_parse_url( $homeUrl );
+			$urlParts['path'] = trailingslashit( Obj::prop( 'path', $urlParts ) ) . $post->post_name . '/';
```
dgwatkins: @strategio what happens if the permalink structure doesn't have a trailing slash?

or if its default permalink structure ...

src/Hooks/LandingPages.php
strategio: Why would it be better @dgwatkins?

```php
function get_post_type( $post = null ) {
	$post = get_post( $post );
	if ( $post ) {
		return $post->post_type;
	}

	return false;
}
```

src/Hooks/LandingPages.php
strategio: >@strategio what happens if the permalink structure doesn't have a trailing slash?

`trailingslashit` will always add a trailing slash @dgwatkins (maybe you are mixing with `usertrailingslashit`), see below:

```php
function trailingslashit( $string ) {
	return untrailingslashit( $string ) . '/';
}
```

>or if its default permalink structure ...

I didn't test this... what I did is just transposing what Elementor is doing:

```php
return get_home_url() . '/' . $post->post_name . '/';
```

But I will test again (maybe Elementor does not allow the default permalink structure).

strategio: Good catch @dgwatkins, indeed Elementor's link filter is not loaded with default permalinks (I added a fix commit 
7dda8b7).

Then, I overlooked Elementor filter and noticed it filters only "published" landing pages, so I added one extra commit for this 82308c3.

src/Hooks/LandingPages.php
dgwatkins: Because we use the WP function designed for this purpose.

If you think its fine, just ignore me. I think of it as forward compatibility :-)

dgwatkins (dismissed):
:+1: @strategio

strategio:
@OnTheGoSystems/compatibility, please one more review.

kkarpieszuk (dismissed):

PR #151, opened 2021-08-06
Improve `DataConvert::unserialize` to be tolerant with already unserialized data

strategio:
This happens with imported templates provide by Astra theme.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8775

Added an extra commit to fix a non-related unit test.

### DIFF
--- a/src/DataConvert.php
+++ b/src/DataConvert.php
@@ -2,6 +2,8 @@
 
 namespace WPML\PB\Elementor;
 
+use WPML\FP\Obj;
+
 class DataConvert {
 
 	/**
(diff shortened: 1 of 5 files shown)


dgwatkins (dismissed):
Everything is perfect :+1:

strategio:
@dgwatkins, I added one more commit, please check again.

Our PB code has some flaw, the "unserialize" code can receive both single or multiple post meta, it ends up decoding the JSON string. And that make impossible to know what we are converting when we receive a full multi-dimensional array... So my last commit is improving the code that is directly affecting us, even though there would be more to fix including in the main PB package (but then it's risky now).

dgwatkins (approved):
@strategio it still looks good

strategio:
I close as "won't fix" for now.

